Issue #173, opened 2015-01-21
Extra packets sent with "-L" option


morphyno:
I'm always noticing an "extra" packet sent when using the "-L" option. Also, this packet is always an malform packet. 

sudo /usr/local/bin/tcpreplay --netmap -i eth6 -L 1000 -tK rx_stc_sig.pcap
Switching network driver for eth6 to netmap bypass mode... done!
File Cache is enabled
Actual: 1001 packets (60060 bytes) sent in 0.000449 seconds.
Rated: 133763919.8 Bps, 1070.11 Mbps, 2229398.66 pps
Flows: 1 flows, 2227.17 fps, 1000000 flow packets, 0 non-flow
Statistics for network device: eth6
        Attempted packets:         1001
        Successful packets:        1001
        Failed packets:            0
        Truncated packets:         0
        Retried packets (ENOBUFS): 0
        Retried packets (EAGAIN):  0
Switching network driver for eth6 to normal mode... done!

I'm seeing this with and without netmap 

sudo /usr/local/bin/tcpreplay -i eth6 -L 1000 -tK rx_stc_sig.pcap
File Cache is enabled
Actual: 1001 packets (60060 bytes) sent in 0.001041 seconds.
Rated: 57694524.4 Bps, 461.55 Mbps, 961575.40 pps
Flows: 1 flows, 960.61 fps, 1000000 flow packets, 0 non-flow
Statistics for network device: eth6
        Attempted packets:         1001
        Successful packets:        1001
        Failed packets:            0
        Truncated packets:         0
        Retried packets (ENOBUFS): 0
        Retried packets (EAGAIN):  0

This is what that packet looks like on wireshark 

![malformed](https://cloud.githubusercontent.com/assets/3343500/5842966/f33b79cc-a15d-11e4-8d97-d3839195cb5b.png)

This is not a big deal or anything, but i'm actually generating Spirent Testcenter signatures from ixgbe (as a source to testcenter) and its confusing Testcenter

fklassen:
Thanks for reporting. I hope to schedule shortly.

fklassen:
Seems to be working now due to lots of bug fixes. Thanks for reporting.

```
fklassen@debian-fred:~/git/tcpreplay$ sudo src/tcpreplay -i eth0 -Kt -L 1000 -l 10 v6.pcap 
File Cache is enabled
Actual: 1000 packets (159984 bytes) sent in 0.603071 seconds.
[...]

Issue #172, opened 2015-01-14
4.1 max bitrate on loop


lancasrj:
As of version 4.1.0 running the following command:
tcpreplay -i eth0 -l 0 capture.pcap 

The first time through everything plays as expected.  Upon looping the bitrate jumps to the maximum rate for the interface and the following message is repeated in the log:

Warning: Packet #9513 has gone back in time!
Warning: Packet #9514 has gone back in time!
Warning: Packet #9515 has gone back in time!
Warning: Packet #9516 has gone back in time!

rdiggz:
Hello, 

Im new to tcpreplay but i have been given the task to get 8 captures to play on CentOS 6.6. I am having this same issue. If I play the captures on Ubuntu server 14.04, i do not get the error. There is also a big CPU usage difference between the 2 but thats another issue. 

I was check to see if there is any update on this?

Thank you for your time and great work!

fklassen:
@abiryukov you are correct. This was fixed in #85. Fixed in version 4.1

rafafade:
Hello 

Sorry if this is mark as closed and that I am missing something, but I have the same problem of  "Warning: Packet #XXXX has gone back in time!" with v4.1.0

The command I execute:
tcpreplay --pid -i eth1 --loop=2 capture_UDP_9760only.pcap

About the .pcap file:
-The packets were captured and filtered with Wireshark
-Then I changed source MAC and IP using tcprewrite
-It contains only UDP traffic (dst port 9760), destination is broadcast (IP and MAC)

tcpreplay version:
[root@localhost pcap_traces]# tcpreplay -V
tcpreplay version: 4.1.0 (build git:v4.1.0)
Copyright 2013-2014 by Fred Klassen <tcpreplay at appneta dot com> - AppNeta Inc.
Copyright 2000-2012 by Aaron Turner <aturner at synfin dot net>
The entire Tcpreplay Suite is licensed under the GPLv3
Cache file supported: 04
Not compiled with libdnet.
Compiled against libpcap: 1.4.0
64 bit packet counters: enabled
Verbose printing via tcpdump: enabled
Packet editing: disabled
Fragroute engine: disabled
Injection method: PF_PACKET send()
Not compiled with netmap

Running under CentOS 6:
[root@localhost pcap_traces]# cat /etc/redhat-release 
CentOS release 6.7 (Final)

Thanks in advance

Rafael

fklassen:
I'll reopen this issue again so that we can review.

[...]

PR #175, opened 2015-01-29
Add option for limiting transmit by time

dexteradeus:
Add --duration option which lets the user set the number of seconds to  transmit
Implemented similarly to the -L option to break out to the send_packets loop and set the abort flag once we have exceeded the time limit.

### DIFF
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -467,6 +467,7 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
     int datalink = options->file_cache[idx].dlt;
     COUNTER skip_length = 0;
     COUNTER start_us;
+    COUNTER end_us;
     uint32_t iteration = ctx->iteration;
     bool unique_ip = options->unique_ip;
     bool preload = options->file_cache[idx].cached;
@@ -474,6 +475,11 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
             (options->speed.mode == speed_mbpsrate && !options->speed.speed);
 
     start_us = TIMEVAL_TO_MICROSEC(&ctx->stats.start_time);
+
+    if (options->limit_time > 0)
+        end_us = start_us + SEC_TO_MICROSEC(options->limit_time);
+    else
+        end_us = 0;
 
     if (options->preload_pcap) {
         prev_packet = &cached_packet;
@@ -494,6 +500,16 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
         if (limit_send > 0 && ctx->stats.pkts_sent >= (COUNTER)limit_send) {
             ctx->abort = true;
             break;
+        }
+
+        /* stop sending based on the duration limit*/
+        if (end_us > 0) {
+            if (gettimeofday(&now, NULL) < 0)
+                errx(-1, "gettimeofday() failed: %s",  strerror(errno));
+            if (TIMEVAL_TO_MICROSEC(&now) > end_us) {
+                ctx->abort = true;
+                break;
+            }
         }
 
         packetnum++;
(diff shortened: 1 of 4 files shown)


seglberg:
:+1:

nuxi:
:+1:

fklassen:
Looks good. Thanks for the much-needed enhancment.

This change is in the area of code as #172 and #173 so I want to test all 3 changes at once. I will schedule for the next beta release.

fklassen:
Hi @dexteradeus. I am doing some testing, and I notice that this option slows things down by about 200Mbps on 10GigE. I think there is a way around it. I'll investigate.

```
root@portwell29:~# /tmp/tcpreplay -i eth7 -Kt --loop 800 /root/smallFlows.pcap
File Cache is enabled
Actual: 11408800 packets (7373224800 bytes) sent in 6.04 seconds.
Rated: 1146774000.0 Bps, 9174.19 Mbps, 1774435.97 pps
Flows: 1209 flows, 188.03 fps, 11394400 flow packets, 14400 non-flow
Statistics for network device: eth7
	Attempted packets:         11408800
	Successful packets:        11408800
	Failed packets:            0
	Truncated packets:         0
	Retried packets (ENOBUFS): 0
	Retried packets (EAGAIN):  0
root@portwell29:~# /tmp/tcpreplay -i eth7 -Kt --duration 5 --loop 800 /root/smallFlows.pcap
File Cache is enabled
Actual: 8647743 packets (5588830473 bytes) sent in 5.00 seconds.
Rated: 1117765800.0 Bps, 8942.12 Mbps, 1729548.25 pps
Flows: 1209 flows, 241.79 fps, 8645501 flow packets, 10926 non-flow
Statistics for network device: eth7
	Attempted packets:         8647743
	Successful packets:        8647743
	Failed packets:            0
	Truncated packets:         0
	Retried packets (ENOBUFS): 0
	Retried packets (EAGAIN):  0
```

src/send_packets.c
```diff
@@ -496,6 +502,16 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
             break;
         }
 
+        /* stop sending based on the duration limit*/
+        if (end_us > 0) {
+            if (gettimeofday(&now, NULL) < 0)
+                errx(-1, "gettimeofday() failed: %s",  strerror(errno));
+            if (TIMEVAL_TO_MICROSEC(&now) > end_us) {
```
fklassen: `gettimeofday()` only fails if there is a parameter problem. No need to check for errors when passing `NULL`

src/tcpreplay_opts.def
```diff
@@ -356,6 +356,19 @@ specify a maximum number of packets to send.
...
 };
 
+flag = {
+    name        = duration;
+    arg-type    = number;
+    max         = 1;
+    arg-default = -1;
+    arg-range   = "1->";
+    descrip     = "Limit the trasmit duration";
+    doc         = <<- EOText
```
fklassen: Typo. `trasmit` should be `transmit`.  Also, units are not specified in short help so possibly better to say `Limit the number of seconds to send`

src/send_packets.c
```diff
@@ -496,6 +502,16 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
             break;
         }
 
+        /* stop sending based on the duration limit*/
+        if (end_us > 0) {
+            if (gettimeofday(&now, NULL) < 0)
```
fklassen: I found the follow patch prevents slowdowns on 10GigE

```
diff --git a/src/send_packets.c b/src/send_packets.c
index cb34caa..e8a8bc6 100644
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -503,7 +503,7 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
         }
 
         /* stop sending based on the duration limit*/
-        if (end_us > 0) {
+        if (!do_not_timestamp && end_us > 0) {
             if (gettimeofday(&now, NULL) < 0)
                 errx(-1, "gettimeofday() failed: %s",  strerror(errno));
             if (TIMEVAL_TO_MICROSEC(&now) > end_us) {
```

fklassen:
FYI, I did some updates. See #223

PR #179, opened 2015-04-30
Fixes an issue with Rated pkt stats overflowing.

sethhall:
It looks like the code for outputting the stats could stand to
be refactored a bit, but this fix kicks the can down the road
a little futher.


sethhall:
This should be ready to be merged

PR #204, opened 2015-10-26
tcprewrite: Handle loopback for linux cooked capture files (Linux SLL)

corentin-p1:
When capturing network packets with tcpdump using the "all" pseudo-interface, the resulting capture file is in linux cooked format (linux SLL). Furthermore, some packets might be coming from the loopback device depending on the activity of the machine. In such a case, when trying to use tcprewrite to replace the linux SLL header with an ethernet header, the program fails with the following error:
```
$ tcprewrite --dlt=enet --enet-dmac=aa:bb:cc:00:00:00 --enet-smac=dd:ee:ff:00:00:00 -i cooked.pcap -o ether.pcap

Fatal Error in tcpedit.c:tcpedit_packet() line 114:
 From ./plugins/dlt_linuxsll/linuxsll.c:dlt_linuxsll_decode() line 201:
DLT_LINUX_SLL pcap's must contain only ethernet packets
```
This pull request address this issue.


fklassen:
Thanks for contributing :+1:

PR #205, opened 2015-10-31
Initial tuntap device support

demirten:


fklassen:
Hi, thanks for the PR.

I notice that the URL points to OS X implementations on tun/tap. I notice that the feature uses /dev/tunX and /dev/tapX. That's the same as Linux tun/tap. Is the feature intended for only OS X, or should it include BSD and Linux?

demirten:
I've tested it under Linux. OS X and FreeBSD parts are based on my previous experience.
Under OS X and FreeBSD, you have to open tun/tap device with exact path like /dev/tap0. Linux has different setup procedure, you have to open /dev/net/tun controlling device and set interface name with the help of an ioctl later.

I'll try to test this on an OS X and FreeBSD.

demirten:
I found that a little extra effort needed on FreeBSD build (on configure stage). I fixed it and now works in FreeBSD too. I'll check Mac OS X to and submit another pull request later. Now I'm closing this.

fklassen:
Thanks Murat. Please put PR into [appneta/tcpreplay/Enhancement_206_tun_tap_support](https://github.com/appneta/tcpreplay/tree/Enhancement_206_tun_tap_support) instead of master. I can test in a branch, then will merge.

Great feature. I was thinking of doing the same to test a tunnel server I wrote and use in production.

PR #207, opened 2015-11-03
Tap device support for Linux and FreeBSD

demirten:
Tap devices must be up and running before using with tcpreplay.

Tap support in Apple devices works a little bit different and it is not compatible with main flow of tcpreplay.

Patch tested in Linux and FreeBSD with different capture files.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -478,6 +478,25 @@ if test x$inet_addr = no ; then
 fi
 
 dnl #####################################################
+dnl Checks for tuntap device support
+dnl #####################################################
+have_tuntap=no
+AC_ARG_ENABLE([tuntap],
+  AS_HELP_STRING([--disable-tuntap], [Disable tuntap support]), [:],
+  [case "$build_os" in
+    linux*)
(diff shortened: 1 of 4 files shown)


fklassen:
Hi Murat. From your comments, can I assume that OS X tun/tap is not supported?

I looked at the code and it looks good to me. I will want to test before merging, but will not be able to work on Tcpreplay until Nov. 23.

demirten:
Yes, I wrote tuntap support for OS X but later I have to remove it.

You need to open /dev/tapX char device and set interface status up in OS X. But all of the procedure must be done within same process. On the other hand, tcpreplay wants to see interfaces on startup, it must be configured before tcpreplay.

So I couldn't find any way to configure tap interface in one process and use it in another process under OS X. If this is possible, adding OS X support will be same as FreeBSD.

fklassen:
Had a chance to play with this, so I merged. Thanks again for contribution.

fklassen:
@demirten I added some comments. If you feel so included, please make updates and send another PR. Tnx, Fred.

PR #208, opened 2015-11-04
add multithreading support

teknoraver:
add option '-a' to spawn multiple threads

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -605,6 +605,7 @@ else
     fi
     if test "$pthread_found" = "yes"; then
         LIBS="$LIBS -pthread"
+        AC_DEFINE([HAVE_PTHREADS], [1], [Do we have phtreads?])
     fi
 fi
 
--- a/configure.ac
+++ b/configure.ac
@@ -603,6 +603,9 @@ else
     if test "$numa_found" = "yes" && test "$pthread_found" = "yes" && test -f "${foundpcap}/../lib/libpfring.a"; then
         pf_ring_found=yes
     fi
+    if test "$pthread_found" = "yes"; then
+        LIBS="$LIBS -pthread"
+    fi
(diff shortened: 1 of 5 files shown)


fklassen:
Hi @teknoraver thanks for the PR. A couple questions before I review and test.

Have you tried compiling on other platforms? I don't see any #ifdef statements, and I am quite sure that pthreads is only available on Linux. Compiling on OS X, FreeBSD etc. will probably fail.

Also, in the past I have found that multiple threads usually performs more poorly than single threaded. I have a 1.2Ghz ARM processor that is able to saturate GigE. Is there a purpose for multi-threading that I am not aware of?

Fred.

teknoraver:
Hi @fklassen 

I have added some ifdefs and configure checks to avoid breaking other platforms, a new pull request is coming

fklassen:
Thanks @teknoraver.

When you make the new Pull Request, can you target [appneta/tcpreplay/Enhancement_208_pthreads](https://github.com/appneta/tcpreplay/tree/Enhancement_208_pthreads) branch instead of master? I want to do some testing for performance. 

My experience at 10GigE+ has not been good with multi-threading. Have you tried it at higher speeds? Is the reason for this development centered around speed, or for added flexibility?

teknoraver:
I'm using tcpreplay on a dual core 400 MHz MIPS gbit system and to reach high speed I used to spawn two tcpreplay instances in parallel.
I will rebase the pull request on appneta/tcpreplay/Enhancement_208_pthreads

fklassen:
Is the idea to spawn a thread per interface? If so, that may be beneficial.

teknoraver:
the idea was to spawn a thread per CPU, obviously you see benefit only with slow CPU and fast interfaces

PR #247, opened 2016-05-12
fragroute fix to support fragmentation of IP datagrams that contain IP Total Length a 0 in header

dsrana1212:
This fix is to support the fragmentation of  IP datagrams generated by wireshark on windows systems that have TCP segmentation offloading to NIC enabled. In such systems, the IP fragmentation is done by NIC and the IP layer just passes the application data buffer to NIC and sets the IP Total Length field in header is set to 0. Since wireshark sniffs packets before they are properly fragmented by NIC, all these IP datagrams will have large payload with IP Total Length header field set to 0.

The fragroute code does fragmentation based on the IP Total Length field value and break the datagram in smaller fragments as specified by user. However, in current case,since IP total length is 0, the fragroute does not fragment the datagram even though the datagram contains huge payload.

Fix :  check if the IP Total Lenght field is zero. If so then read the length of datagram from the incoming buffer length.

Tests : tested with wireshark captures containing valid IP Total Length and also with captures containing zero IP Total Length.


fklassen:
Thanks for the PR.

dsrana1212:
No problem. This tool hass helped me a lot at my job:). Its time to give back:)

PR #256, opened 2016-08-08
give user CFLAGS precedence

GabrielGanne:
Hi,
This is a small one-liner changing the CFLAGS order, so that we can override -O3.
BR.


fklassen:
Thanks for the PR.

Issue #190, opened 2015-06-24
Tcprewrite: UDP checksum calculated wrong for fragmented UDP packet


sorenhede:
When using -C on pcap-file containing fragmented UDP packets, the UDP checksum is incorrect.

It is possible to reproduce the issue using the file "http://loki.stockton.edu/~olanm/csis4222/homework/fragments.pcap" 
With:
tcprewrite -C --infile=fragments.pcap --outfile=output.pcap

View of the fragments.pcap file in Wireshark
![fragments](https://cloud.githubusercontent.com/assets/10185732/8332106/5a441a82-1a8c-11e5-922d-0879365930d6.png)

View of the output.pcap with incorrect checksum after tcprewrite -C in Wireshark
![output](https://cloud.githubusercontent.com/assets/10185732/8332111/5e210872-1a8c-11e5-8f2d-02e893173329.png)

My current tcprewite version:
tcprewrite version: 4.1.0 (build git:v4.1.0)
Copyright 2013-2014 by Fred Klassen <tcpreplay at appneta dot com> - AppNeta Inc.
Copyright 2000-2012 by Aaron Turner <aturner at synfin dot net>
The entire Tcpreplay Suite is licensed under the GPLv3
Cache file supported: 04
Not compiled with libdnet.
Compiled against libpcap: 1.5.3
64 bit packet counters: enabled
Verbose printing via tcpdump: enabled
Fragroute engine: disabled

fklassen:
Hi, any chance that you have UDP checksum offload enabled on the NIC? 

I'll schedule some time to look into this for next beta release.

sorenhede:
Hi, the fragments.pcap testfile used in this example have correct UDP checksum - but when pass it thrugh tcprewrite with -C the UDP checksum become incorrect in the output file.

I believe there might be a problem with calculation of fragmented UDP checksum

fklassen:
It appears that #225 fixes this bug, mostly.

The underlying issue is that that `tcpreplay-edit` and `tcprewrite` can only process one packet at a time. For many operations, if something is edited, packet checksums must be updated. Original technique was to recalculate the entire checksum. This doesn't work for fragments, which require checksumming frames that will be sent/rewritten in future frames.

Fix includes implementing [RFC 1624 Incremental Checksum](https://tools.ietf.org/html/rfc1624). It combines local edit differences and then applies it to the already calculated checksum.

This means that if edits are done to only the first frame of a fragmented packet, then everything works. Those options include...
[...]

PR #258, opened 2016-08-22
Add data fuzzing option to tcpreplay-edit

GabrielGanne:
Hi,

This is a proposal to add high-level fuzzing to tcpreplay-edit.
I am using it mostly to test protocols over http with pcap files.

I named the option --fuzz-seed because there was a --seed option already used for IP randomization.
Tell me if you do not think this is appropriate.

I am emulating a packet drop by setting the packet len to 0.
For this to work, I had to change some assertions that it could not be null (as in dlt_en10mb_l2len). I am not sure this is a legitimate change.

Regards,


fklassen:
Thanks for the PR.

I'll have to review the change when I get back on this project. I hope to work on the next release in Sept.

src/tcpedit/plugins/dlt_en10mb/en10mb.c
```diff
@@ -586,22 +586,34 @@ dlt_en10mb_get_mac(tcpeditdlt_t *ctx, tcpeditdlt_mac_type_t mac, const u_char *p
 int
 dlt_en10mb_l2len(tcpeditdlt_t *ctx, const u_char *packet, const int pktlen)
 {
+    int l2len;
     struct tcpr_ethernet_hdr *eth = NULL;
     
     assert(ctx);
     assert(packet);
-    assert(pktlen);
+
```
fklassen: This mode is acceptable if done to all 13 plugins. Currently you have only modified Ethernet. 

Once this change is made, the PR looks good to me.

GabrielGanne:
Hi,

Thanks for reviewing.

I've started to change the asserts on pktlen on all the plugins. Seeing that I'll have to patch all of the following, it will generate quite some diff, so please be patient.
``` bash
git grep -c  assert.*pktlen -- src/tcpedit/plugins/
src/tcpedit/plugins/dlt_en10mb/en10mb.c:5
src/tcpedit/plugins/dlt_hdlc/hdlc.c:7
src/tcpedit/plugins/dlt_ieee80211/ieee80211.c:7
src/tcpedit/plugins/dlt_ieee80211/ieee80211_hdr.c:1
src/tcpedit/plugins/dlt_jnpr_ether/jnpr_ether.c:7
src/tcpedit/plugins/dlt_linuxsll/linuxsll.c:7
src/tcpedit/plugins/dlt_null/null.c:7
src/tcpedit/plugins/dlt_plugins.c:5
src/tcpedit/plugins/dlt_pppserial/pppserial.c:7
src/tcpedit/plugins/dlt_radiotap/radiotap.c:5
src/tcpedit/plugins/dlt_raw/raw.c:7
src/tcpedit/plugins/dlt_template/plugin.c.tmpl:7
src/tcpedit/plugins/dlt_user/user.c:7
src/tcpedit/plugins/dlt_utils.c:2
```

Also, I've stumbled upon some code wich I feel uncomfortable changing but fell like I should, such as in :  

src/tcpedit/plugins/dlt_radiotap/radiotap.c
``` c
int
dlt_radiotap_l2len(tcpeditdlt_t *ctx, const u_char *packet, const int pktlen)
{
    uint16_t radiolen;
    assert(ctx);
    assert(packet);
    assert(pktlen);

    /* endianness issue here */
    memcpy(&radiolen, &packet[2], 2); 
    return (int)radiolen;
}
```

If you agree, I propose to leave those untouched within this commit, and to send you another patch later.

fklassen:
@GabrielGanne Thanks for doing this. I will hold up 4.2 release until we get this resolved and tested.

Make all the changes you feel appropriate, and I'll review. I'll fix things if I see anything that concerns me.

GabrielGanne:
@fklassen This should be it.
I separated the commits as I made them, but feel free to squash them as you see fit.

src/tcpedit/edit_packet.c
```diff
@@ -63,7 +63,9 @@ fix_ipv4_checksums(tcpedit_t *tcpedit, struct pcap_pkthdr *pkthdr, ipv4_hdr_t *i
     
 
     /* calc the L4 checksum if we have the whole packet && not a frag or first frag */
-    if (pkthdr->caplen == pkthdr->len && (htons(ip_hdr->ip_off) & IP_OFFMASK) == 0) {
+    if (pkthdr->caplen == pkthdr->len
+            && pkthdr->len > ntohs(ip_hdr->ip_len)
+            && (htons(ip_hdr->ip_off) & IP_OFFMASK) == 0) {
```
fklassen: Not sure I understand the logic of `&& pkthdr->len > ntohs(ip_hdr->ip_len)`. Can there be conditions where the packet length is altered but ip_hdr->ip_len is not? If so, wouldn't it be better to not set `tcpedit->fixcsum` / `needtorecalc` so that this function is not called?

Also, what about IPv6 in line 95?

If this condition remains, we should at least do `(ntohs(ip_hdr->ip_len) + (ip_hdr->ip_hl << 2))` on the right side.

Note that there will be a merge conflict here due to #190. I'll fix that downstream.

src/tcprewrite.c
```diff
@@ -229,7 +229,7 @@ rewrite_packets(tcpedit_t *tcpedit, pcap_t *pin, pcap_dumper_t *pout)
     COUNTER packetnum = 0;
     int rcode;
 #ifdef ENABLE_FRAGROUTE
-    int frag_len, i, proto;
+    int frag_len, proto;
```
fklassen: This will break compile if using `./configure --enable-debug` and when using FRAG_ROUTE. I have been working on a fix in 4.2.

src/tcpedit/edit_packet.c
GabrielGanne: The comment following `needtorecalc` is that it should be set should the packet change, so it feels right to set the flag when I edit the packet bytes or length through fuzzing.
Do you agree ?

I think I messed up that part: the test should fail if `ip_hdr->ip_len < (iphdr_->ip_hl << 2) // ie. totlen < hdrlen` (right ?)
I'll rework that patch, and also have a look at ipv6 to be coherent.

Also, I'll rebase so as to prevent merge conflicts.

src/tcprewrite.c
GabrielGanne: True.
Had not seen that.
I'll remove the patch.
Sorry.

fklassen:
@GabrielGanne if you rebase to branch `Feature_302_fuzz_option` I can do updates to it, and merge into a beta branch. If you have questions, let me know.

src/tcpedit/edit_packet.c
fklassen: Agreed. I'll watch for your patch. Also please make PR to branch `appneta/tcpreplay/Feature_302_fuzz_option`

fklassen:
Hi @GabrielGanne. I'll have a closer look at this tonight. If you are OK with it, I'll merge the PR and with it in it's own branch, we can both work on this together.

Note that I just added a couple plugins that are failing. The codacy  plugin is new and needs some configuration to be less noisy. However it does point out that there are 2 consecutive breaks in a switch statement.

More interesting is travis-ci pointing out that there is a test failure on MTU. This may or may not be an actual test failure. But it may be nice to add a test for fuzz to make sure it behaves properly.

This is a big feature so thanks for hanging in there and helping get things production ready.

GabrielGanne:
Hi,

Sure, I'll add a test case for the fuzzing.

I had a look at the mtu diff, and it turns out that the diff is the tcp checksum of packets which are not changed by the new mtu, but had an invalid checksum before.
My guess is that this is a bug.
I'll try to fix it tonight.

fklassen:
Thanks. I'll leave this PR open until you feel it is time to merge. I'll do some heavy testing on it once it is merged.

GabrielGanne:
@fklassen 
I updated my branch :
I fixed the regression on the mtu test which was an error from my part.
I added a very simple test case for l7 fuzzing.
I also believe I fixed the issues you pointed out in your previous review.

I did some tests on big pcaps, and the fuzzing is pretty noisy because of all the TCPEDIT_WARN values.
As far as I can tell, they all were valid, but not very interesting when fuzzing is enabled.
I do not see a way to get rid of them.

Do tell me if you wish me to rebase, or squash certain commits.

fklassen:
Great work @GabrielGanne. I'll merge to my feature branch and start testing. If you have any other changes, just PR into the same feature branch.

Thanks for hanging in there and getting it done. Fred.

PR #259, opened 2016-08-23
declare the headers generated by autoopts as BUILT_SOURCES

GabrielGanne:
Hi,

This patches adds the *.h counterpart of all *.def files of src/Makefile.am as BUILT_SOURCES.
This way they should be generated before (for example) tcpreplay_api.c.

This should allow parallel build (It works for me, but I do not know how to test if I missed a target).

Regards,


fklassen:
Thanks for the PR. I expect to work on this around the end of September.

PR #263, opened 2016-10-29
initial travis-ci support

chipitsine:
please add this project to https://travis-ci.org

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+sudo: required
+dist: trusty
+
+os: linux
+
+language: c
+
+compiler:
+  - gcc
+  - clang
+
+addons:
+  apt:
+    packages:
+    - libpcap-dev
+
+cache:
+  apt: true
+
+script:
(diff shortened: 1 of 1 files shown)


fklassen:
Thanks for the PR.

I am a bit worried about the rights I give to Travis CI. It seems I have to give the world to all repos, including private ones. We trust GitHub with our private repos, but don't have that trust with Travis.

I sent an email to GitHub support. They say Travis gets right to all repos, both public and private. They don't have access to private sources. GH does not vouch for vendors who I give permissions to. So, risk is on my end.

I will hold onto this PR until I can do more investigation.

chipitsine:
according to https://docs.travis-ci.com/user/github-oauth-scopes/ travis-ci itself is granted a very limited subset of permissions.

what exactly are you afraid of ?

fklassen:
After talking to GH Support, realize that Travis has too much access to my private repos if I use my primary account. I set up my secondary account to have `outside collaborator` access. This allows me to give Travis access to only Tcpreplay and no other repos.

PR #268, opened 2016-12-02
Add protection against packet drift.

guscott:
Hi,

Currently, the next packet sent time is based on the previous packet. When a lot of packets are sent, a small jitter may be inserted and all subsequent packet will have the same added jitter; tcpreplay will not recover from this jitter. 
The proposed improvement add a comparison with the system time to recover from those timestamp drift. It also avoid updating the system time and sleeping when there is a jitter to recover from.

Regards.

### DIFF
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -69,7 +69,7 @@ extern tcpedit_t *tcpedit;
 extern int debug;
 #endif
 
-static void calc_sleep_time(tcpreplay_t *ctx, struct timeval *pkt_time,
+static bool calc_sleep_time(tcpreplay_t *ctx, struct timeval *pkt_time,
         struct timeval *last, COUNTER len,
         sendpacket_t *sp, COUNTER counter, timestamp_t *sent_timestamp,
         COUNTER *start_us, COUNTER *skip_length);
@@ -470,7 +470,7 @@ preload_pcap_file(tcpreplay_t *ctx, int idx)
 void
 send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
 {
-    struct timeval print_delta, now;
+    struct timeval print_delta, now, first_pkt_ts, pkt_ts_delta;
     tcpreplay_opt_t *options = ctx->options;
     COUNTER packetnum = 0;
     COUNTER limit_send = options->limit_send;
@@ -495,6 +495,7 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
 
     ctx->skip_packets = 0;
     start_us = TIMEVAL_TO_MICROSEC(&ctx->stats.start_time);
+    timerclear(&first_pkt_ts);
 
     if (options->limit_time > 0)
         end_us = start_us + SEC_TO_MICROSEC(options->limit_time);
@@ -514,6 +515,7 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
     while (!ctx->abort &&
             (pktdata = get_next_packet(ctx, pcap, &pkthdr, idx, prev_packet)) != NULL) {
 
+        now_is_now = false;
         packetnum++;
 #if defined TCPREPLAY || defined TCPREPLAY_EDIT
         /* do we use the snaplen (caplen) or the "actual" packet len? */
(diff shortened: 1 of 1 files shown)


fklassen:
Thanks for the PR.  I had a quick glance but will dig deeper when I get a chance.

Have you tested at 10GigE + rates? My only concern is whether `gettimeofday()` is called more aggressively. If so, it will affect performance. Getting a timestamp is very expensive.

guscott:
The change actually call gettimeofday() less often than before. So, it will result in a performance boost.

I don't have a 10GigE NIC available to test, but I tested with a 1G NIC and small VoIP packets. Before the PR, tcpreplay was struggling to output ~350+Mbps in realtime. Now, tcpreplay is able to output ~600Mbps  in realtime without any issue. I have not tested with higher bandwidth, but it should be able to drive at line rate if the processor running tcpreplay is strong enough.

src/send_packets.c
```diff
@@ -560,35 +562,54 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
...
              */
+            dbgx(4, "This packet time: " TIMEVAL_FORMAT, pkthdr.ts.tv_sec, pkthdr.ts.tv_usec);
             skip_length = 0;
             ctx->skip_packets = 0;
-            now_is_now = true;
-            gettimeofday(&now, NULL);
 
             /*
-             * Only sleep if we're not in top speed mode (-t)
+             * Only sleep if we're not in top speed mode (-t) or late
```
fklassen: Do you mean `or not late` ?

src/send_packets.c
```diff
@@ -560,35 +562,54 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
...
              *
              * This also sets skip_length which will avoid timestamping for
              * a given number of packets.
              */
-            calc_sleep_time(ctx, &pkthdr.ts, &ctx->stats.last_time, pktlen, sp, packetnum,
-                    &ctx->stats.end_time, &start_us, &skip_length);
+            timersub(&pkthdr.ts, &first_pkt_ts, &pkt_ts_delta);
+            if (calc_sleep_time(ctx, &pkt_ts_delta, &ctx->stats.last_time, pktlen, sp, packetnum,
+                    &ctx->stats.end_time, &start_us, &skip_length))
+            {
```
fklassen: Slight request for style change to be consistent with project: brace moved up a line.

src/send_packets.c
```diff
@@ -806,35 +820,54 @@ send_dual_packets(tcpreplay_t *ctx, pcap_t *pcap1, int cache_file_idx1, pcap_t *
         if (options->flow_stats && !options->file_cache[cache_file_idx].cached)
             update_flow_stats(ctx, sp, pkthdr_ptr, pktdata, datalink);
 
+        if (ctx->first_time)
+        {
```
fklassen: Same style change request.

src/send_packets.c
```diff
@@ -806,35 +820,54 @@ send_dual_packets(tcpreplay_t *ctx, pcap_t *pcap1, int cache_file_idx1, pcap_t *
...
              *
              * This also sets skip_length which will avoid timestamping for
              * a given number of packets.
              */
-            calc_sleep_time(ctx, &pkthdr_ptr->ts, &ctx->stats.last_time, pktlen, sp, packetnum,
-                    &ctx->stats.end_time, &start_us, &skip_length);
+             timersub(&pkthdr_ptr->ts, &first_pkt_ts, &pkt_ts_delta);
+            if (calc_sleep_time(ctx, &pkt_ts_delta, &ctx->stats.last_time, pktlen, sp, packetnum,
+                    &ctx->stats.end_time, &start_us, &skip_length))
+            {
```
fklassen: Same style change: move brace up a line

src/send_packets.c
guscott: Yes, I rephrased the comment.

src/send_packets.c
guscott: done

src/send_packets.c
guscott: done

src/send_packets.c
```diff
@@ -69,7 +69,7 @@ extern tcpedit_t *tcpedit;
 extern int debug;
 #endif
 
-static void calc_sleep_time(tcpreplay_t *ctx, struct timeval *pkt_time,
+static bool calc_sleep_time(tcpreplay_t *ctx, struct timeval *pkt_time,
```
guscott: I added a return value to bypass the update of 'now' to reduce the number of call to gettimeofday()

fklassen: Ah, I see that now. Makes sense.

PR #276, opened 2017-01-31
common/sendpacket: avoid build error with musl-libc

### LINKED ISSUE #275: cross-compilation fail with musl-libc
Hi Fred,

Just reporting in from buildroot with a build failure for tcpreplay when using musl as the libc[1]

````
In file included from /home/peko/autobuild/instance-1/output/host/usr/arm-buildroot-linux-musleabihf/sysroot/usr/include/linux/if_tun.h:20:0,
                 from sendpacket.c:160:
/home/peko/autobuild/instance-1/output/host/usr/arm-buildroot-linux-musleabihf/sysroot/usr/include/linux/if_ether.h:141:8: error: redefinition of 'struct ethhdr'
 struct ethhdr {
        ^
````

According to the musl FAQ[2] the fix is to avoid including linux/if_tun.h. Looks like thats what socat ended up doing [3]. If I get a chance I'll send a PR unless someone else fixes it first.

[1] - http://autobuild.buildroot.net/results/dde/dde54ccefa32af1f46a4af5b6caed822c3209dfc/build-end.log
[2] - http://wiki.musl-libc.org/wiki/FAQ#Q:_why_am_i_getting_.22error:_redefinition_of_struct_ethhdr.2Ftcphdr.2Fetc.22_.3F
[3] - https://github.com/openwrt/packages/pull/1386/commits/a0968fae

cpackham:
Fixes #275

The musl C library has a definition of 'struct ethhdr' that conflicts
with the one provided by linux. Avoid this by not including
netinet/if_ether.h when building with something other than glibc or
uclibc.

Signed-off-by: Chris Packham <chris.packham@alliedtelesis.co.nz>

### DIFF
--- a/src/common/sendpacket.h
+++ b/src/common/sendpacket.h
@@ -26,7 +26,9 @@
 #ifdef __NetBSD__
 #include <net/if_ether.h>
 #else
+#if defined(__UCLIBC__) || defined(__GLIBC__)
 #include <netinet/if_ether.h>
+#endif
 #endif
 
 #ifdef HAVE_QUICK_TX


cpackham:
Hi Fred,

One of the buildroot devs has an alternate fix that you might consider instead https://git.buildroot.net/buildroot/commit/?id=69ced963555b it makes the fix in musl instead of on a per-app basis.

I'm just making this PR as an FYI if someone else building with musl reports a problem. I haven't seen any sign that musl itself has picked up the fix from buildroot.

fklassen:
For now I will close this PR until a proper test for musl can be created. At this time, this change has too much risk for breaking other builds.

PR #281, opened 2017-02-26
Release 4.2

fklassen:


chipitsine:
there's no "gcc" on osx, you can exclude it

fklassen:
@chipitsine thanks. I am trying to work out the OS X build so It will take a few attempts. Right now `libpcap` is difficult to install.

chipitsine:
I'm afraid "compiler: gcc" under osx is actually "clang"

fklassen:
@chipitsine I understand. I am simply making sure that changes work on both the oldest Xcode (6.4) and newest (8.2). I have had issues in the past with various versions of Xcode. I distinguish between the 2 by specifying gcc on old, and clang on new. It makes it easier to see the difference in the summary.

PR #284, opened 2017-02-27
travis-ci: cleanup

chipitsine:
I added that option by mistake, travis is not capable of cacheing apt


fklassen:
Thanks for the PR. This would cause merge issues, so I created PR #285. Will close this PR.

Issue #250, opened 2016-07-04
Feature request - set netmap mode once and don't reset


fklassen:
As per request

Hello
I hope you are doing well.
I looked around, but I didn’t see specific functionality described below.
We need to plays hours of unique PCAP data at 10Gbs.   The complete test suite of PCAPs is more than 40TB.  Our previous 1Gbs environment had tcpreplay iterate through thousands of 2Gb PCAP files, one by one.    With 10Gb and the –netmap option, there is a significant delay and “deadtime” for each PCAP file as tcpreplay shifts from normal to bypass and back to normal every time it is invoked.   The interfaces have no other purpose in life other than tcpreplay and their output is aggregated to get the speeds required for testing.   The output of multiple tcpreplay servers will be combined to produce 40Gbs.  Right now, the small PCAPs and mode transition time produce a “sawtooth” graph, while I’m looking to achieve a straight horizontal line.

Is there a way to permanently put the interface in the bypass mode leave it there, thus avoiding this mode transition time?  

Thanks for any ideas.

epearce2:
I didn't think of this before, but I could potentially list hundreds of different pcaps on the tcpreplay command line so that there is only one 'mode transition' delay involved when playing them.  But I would still like the ability to set the 'mode' of the interface on a permanent basis, perhaps at boot time.

fklassen:
Might be best to do it that way. Not sure when I can get this feature scheduled.

fklassen:
I'll have another look at this. However, are you aware that you can pass multiple PCAP files in the command line parameters? That way you can run many PCAP files back-to-back without the interface going up and down.

fklassen:
Based on [this response to my question to netmap group](https://github.com/luigirizzo/netmap/issues/279#issuecomment-282961082), it appears that it is nearly impossible to do this. It certainly is impractical. Once the program closes, netmap will reset the interface.

Will close as invalid.

fklassen:
Reopen. Asking for a feature in Netmap that would allow it.

fklassen:
based on response https://github.com/luigirizzo/netmap/issues/279 this feature cannot be implemented

PR #291, opened 2017-02-28
#250 implement interface hold feature, but doesn't work

fklassen:


fklassen:
This isn't working. The interface returns to normal mode when the program closes. Since the file handle must be closed, netmap will shut down, then it must be started up again when the program starts.

I will contact Netmap folks about this.

fklassen:
Close until issues in #250 are addressed by `netmap` team.

PR #335, opened 2017-03-18
Remove -O3 from CFLAGS

jer-gentoo:
-O3 can have adverse effects and may result in slower code, so users should be able to set and/or override it in USER_CFLAGS.


fklassen:
@jer-gentoo thanks for the PR. How about substituting  with `-O2` flag? I'm not an expert, but I do see `-O2` on lots of things, including Linux Kernel.

jer-gentoo:
Well, the Linux kernel does normally default to -O2 unless you tell it to use -Os instead. The basic point of the PR is to not force it to begin with, but let the (user) environment decide unless you know very well why you need generic defaults for GCC compiler flags.

fklassen:
👍

PR #334, opened 2017-03-18
Replace deprecated AM_CONFIG_HEADER

jer-gentoo:

PR #336, opened 2017-03-19
Enhancement #334 am config header staging

fklassen:


fklassen:
@jer-gentoo I took the opportunity to modernize as per https://www.gnu.org/software/automake/manual/html_node/Public-Macros.html. I tested, and you are correct. I can pass optimizations to configure, e.g. `./configure CFLAGS=-02`.

Issue #333, opened 2017-03-18
Option --fuzz corrupts packets


fklassen:
The attached capture was created with the test on OS X. Original file is `test.pcap`, which distributes with source code. 

Notice that the packet 15 has changed from an IP packet to an invalid Ethernet II.

Most likely this is because [this TODO](https://github.com/appneta/tcpreplay/blob/Feature_302_fuzz_option/src/tcpedit/fuzzing.c#L81) was never implemented.

[test.rewrite_l7fuzzing1.pcap](https://github.com/appneta/tcpreplay/files/852126/test.rewrite_l7fuzzing1.pcap.txt)

fklassen:
@GabrielGanne can you comment on this? Is it expected that IP packets can be converted to something else?

I may be able to merge to Release_4.2 without this addressed.

GabrielGanne:
Hi @fklassen ,

You are correct: fuzzing can indeed change any Byte, including ethertype.
I'm afraid I had forgotten this TODO ...

This is an idea I had which differs from what we would do at my work (dpi) : since most people do not care about code branches for many different protocols, editing the low-level protocol Bytes would result in dropping the packet instead of changing the course of the packet through the code.

Anyway, this TODO only requires parsing and skipping the first 3 layers, so I should be done quickly enough.
(or if you wish to close this release, I can do it for the next one)

Sorry about the trouble.

fklassen:
@GabrielGanne, not a problem. My preference is to hold off 4.2 to get this feature in.  But if you don't have time to do this, I could release 4.2 (scheduled for Monday) and then release a 4.2.1 when you are ready.

As for updates, feel free to make a PR into my `Feature_302_fuzz_option` branch. Or easier still, just make a PR to my master branch. I found I can edit the PR and push to the branch of my choice.

PR #341, opened 2017-03-20
fuzzing  ignores layers 2, 3, and 4

GabrielGanne:
For example, this means :
- skipping TCP SYN packets
- no ethertype, ip addresses, not port number will be modified

Signed-off-by: Gabriel Ganne <gabriel.ganne@enea.com>

### DIFF
--- a/src/tcpedit/fuzzing.c
+++ b/src/tcpedit/fuzzing.c
@@ -37,8 +37,8 @@ fuzz_get_sgt_size(uint32_t r, uint32_t caplen)
 }
 
 static inline int
-fuzz_reduce_packet_size(tcpedit_t *tcpedit, struct pcap_pkthdr *pkthdr,
-        u_char **pktdata, COUNTER new_len)
+fuzz_reduce_packet_size(tcpedit_t * tcpedit, struct pcap_pkthdr * pkthdr,
+        COUNTER new_len)
 {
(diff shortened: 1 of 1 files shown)


GabrielGanne:
I've had unexpected trouble at home, and this took me way longer than I thought.
Sorry about that.

I've ran some checks tonight, and it now seems stable and functional.
I've tested with `tshark -T fields  -e ip.addr -e tcp.port -r <pcap>` I can only see diffs on dropped packets.

GabrielGanne:
This is a fix for : https://github.com/appneta/tcpreplay/issues/333

src/tcpedit/fuzzing.c
```diff
@@ -60,25 +60,100 @@ fuzz_reduce_packet_size(tcpedit_t *tcpedit, struct pcap_pkthdr *pkthdr,
...
+    datalen -= l3data - *pktdata;
+
+    if (datalen <= 0) {
+        return -1;
+    }
+
+    /* switch on layer 2 */
+    switch (ntohs(tcpedit->dlt_ctx->proto))
+    {
+        /* TODO: ntohs on constants could be done at compile time */
```
fklassen: Good point. It's done that way in some Linux kernel code.

fklassen:
👍

PR #397, opened 2017-05-25
Fix sleep calculation when using multiple pcap files

lostd:
Assume every file starts a new replay session, so reset the time this
session started, the time last packet was sent, and the timestamp of
the first packet in this trace.  Don't reset the global start time so
that stats are correct.

### DIFF
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -512,9 +512,16 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
         prev_packet = NULL;
     }
 
+    if (ctx->first_time) {
+        gettimeofday(&now, NULL);
+        now_is_now = true;
+    }
+
     if (!top_speed) {
         gettimeofday(&now, NULL);
         now_is_now = true;
+        start_us = TIMEVAL_TO_MICROSEC(&now);
+        ctx->first_time = true;
     } else {
         now_is_now = false;
     }
@@ -576,8 +583,7 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
                     options->cache_packets ? sp : NULL, &pkthdr, pktdata, datalink);
 
         if (ctx->first_time) {
-            /* get time and timestamp of the first packet */
-            gettimeofday(&now, NULL);
+            /* get timestamp of the first packet */
             memcpy(&first_pkt_ts, &pkthdr.ts, sizeof(struct timeval));
         }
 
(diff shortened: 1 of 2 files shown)


fklassen:
@lostd thanks for the PR.

I was not aware that there was an issue with multiple files. How do you recreate the issue?

Your patch will break performance. Maybe with some rework the patch can get accepted. I'll add comments inline.

src/send_packets.c
```diff
@@ -498,11 +498,13 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
     bool now_is_now;
 
     ctx->skip_packets = 0;
-    start_us = TIMEVAL_TO_MICROSEC(&ctx->stats.start_time);
+    ctx->first_time = true;
+    gettimeofday(&now, NULL);
```
fklassen: This will cause regression issues because every packet is timestamped. Timestamping is hugely expensive and is only done when absolutely necessary.

Overall, the logic is that timestamps are to be done only when necessary, or if it is safe to do so. Timestamps are hugely expensive, especially at 10GigE + speeds. For example, if it is determined that TX is behind by the designated rate by 10msec on a 10GigE link, the loop will send 1.25MB with no timestamping. This usually allows TX to catch up.

This logic allows me to specify, say 9.5Gbps rate and I get 9.5Gbps. With this mod, I would expect that to drop to about 6-8Gbps.

lostd: Hi @fklassen, it's not every packet that gets a timestamp here, this is outside the "while get_next_packet", so every iteration will get an initial timestamp. In small traces in a loop, I guess this is quite frequent though. Would protecting it with a `!top_speed` make the diff merge-able?

To reproduce: `tcpreplay -i if0 trace1.pcap trace2.pcap` where trace1 != trace2. In my case they were traces split from a bigger pcap file, but I don't think it matters. For trace2 `first_pkt_ts` will be cleared and `start_us` referring to the time trace1 started, so it would try to sleep for a huge amount of time.

lostd:
I believe this new diff does not add any timestamping overhead and is smaller too. Comments?

fklassen:
@lostd having a second look, and I believe your last comment is correct. I rebased to a working branch so I can do some testing in my 10GigE lab.

Once again, thanks for the contribution.

lostd:
Hi! Any chance you get this in a next bugfix release?

fklassen:
@lostd I will be starting work on a new release for mid Sept. We have been in a code freeze over the summer.

Scheduled for 4.2.7.

lostd:
Hi there, do you want me to provide a fresh, rebased pull req?

fklassen:
> Hi there, do you want me to provide a fresh, rebased pull req?

I believe this is fixed. As per #399 this is fixed in #419.

lostd:
Verified the fix with latest code, thanks!

PR #416, opened 2017-08-28
Avoid non-blocking behaviour when using STDIN

dlangille:
While on BSD systems, tcprewrite likely sets `O_NDELAY` which is 
proper for most posix systems as this is the newer flag and they 
likely treat both similarly. However BSD and unix derivatives should 
likely utilize `FIONBIO` due to known issues with reading from tty 
with a 0 byte read returning -1 opposed to 0.

Reference comment 1: https://stackoverflow.com/questions/1150635/unix-nonblocking-i-o-o-nonblock-vs-fionbio

on behalf of Cisco Talos

### DIFF
--- a/src/tcprewrite.c
+++ b/src/tcprewrite.c
@@ -35,6 +35,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/types.h>
+#include <sys/ioctl.h>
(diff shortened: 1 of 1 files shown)


fklassen:
Thanks for the PR. I'll review and schedule for next release.

PR #425, opened 2017-11-19
Add ability to change tcp SEQ/ACK numbers

mariodsantana:
For playing around on networks that drop or log repeated seq/ack numbers.


fklassen:
Thanks for the PR.

It appears that Travis CI had a build failure. Looking at the above logs, it appears that `sudo make test` failed.

mariodsantana:
Closing this pull request until I can figure out why make test is failing.

mariodsantana:
@fklassen - boy, that should have been obvious.  This patch generates a _random_ shift to the SEQ/ACK numbers in the input packets, so the test output will not be the same across multiple test runs.  The test harness seems to require that multiple runs with the same args on the same input will create the same output - with this patch, this will happen approximately once every 2^32 test runs on average.  ;)

Let me know how you'd like to proceed.  I don't have time in the next few weeks to rework the test harness here, but I'd like to give back if feasible.

fklassen:
I usually use un-seeded random numbers for my tests, and make sure to use the random number generator included in the source tree. This ensures tests are consistent across all platforms including Big-Endian machines. It is important to get tests right because we are now in the Debian/Ubuntu tree which will reject versions that fail tests on any platform.

If you don't have time to look at this, I plan to take some vacation time at the end of the year. I use this time to catch up on Tcpreplay bugs and features. I can have a closer look then. I can also test on multiple platforms to make sure nothing breaks.

mariodsantana:
Understood about the testing.  I do use the normal rand() function, since the quality of the randomness isn't important.  I'll see if I can clean it up before you get a chance to look at it during the holidays.

Thanks!

fklassen:
@mariodsantana I merged into a new PR #449 so I can test and prep for 4.3 release. Can you look into 2 bugs? #450  and #451 

Thanks, Fred.

fklassen:
merged to 4.3 beta2 in #449

PR #428, opened 2017-12-04
Ensure the argument to {to,is}xxxx() is within allowed range.

he32:
The allowed value range for the arguments to the toxxxx() and
isxxxx() set of functions or macros is the value range of "unsigned
char" plus -1 (for EOF), using them with signed char args may invoke
undefined behaviour (or out-of-bounds-read); this set of changes
casts the args to unsigned char.

### DIFF
--- a/libopts/makeshell.c
+++ b/libopts/makeshell.c
@@ -396,7 +396,7 @@ emit_usage(tOptions * opts)
 
         /* Copy the program name into the time/name buffer */
         for (;;) {
-            if ((*pzPN++ = (char)tolower(*pz++)) == NUL)
+            if ((*pzPN++ = (char)tolower((unsigned char)*pz++)) == NUL)
                 break;
         }
 
@@ -653,7 +653,7 @@ emit_match_expr(char const * name, tOptDesc * cod, tOptions * opts)
              *  They must not be the same.  They cannot be, because it would
              *  not compile correctly if they were.
              */
-            while (toupper(od->pz_Name[match_ct]) == toupper(name[match_ct]))
+            while (toupper((unsigned char)od->pz_Name[match_ct]) == toupper((unsigned char)name[match_ct]))
                 match_ct++;
 
             if (match_ct > min_match_ct)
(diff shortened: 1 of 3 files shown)


fklassen:
Hi, is this fixing a particular bug seen in Tcpreplay? If so, the fix should be submitted to (autogen)[https://www.gnu.org/software/autogen/]. Tcpreplay supplies the `libopts` source as a convenience because it isn't readily available on most systems. It was obtained by the following command:

    gunzip -c `autoopts-config libsrc` |  tar -xvf -

Also, according to Linux `man tolower` the cast should be to `int` not `unsigned char`.

he32:
> Hi, is this fixing a particular bug seen in Tcpreplay? If so,
> the fix should be submitted to (autogen)
> [https://www.gnu.org/software/autogen/]. Tcpreplay supplies the
> `libopts` source as a convenience because it isn't readily
> available on most systems. It was obtained by the following
> command:
>
>     gunzip -c `autoopts-config libsrc` |  tar -xvf -

Hmm, ok, I'll look into doing that.

> Also, according to Linux `man tolower` the cast should be to
> `int` not `unsigned char`.

The NetBSD man page ctype(3) agrees that the prototype for these
functions take an int as argument.  However, it also contains
this text:

CAVEATS
     The first argument of these functions is of type int, but only a very
     restricted subset of values are actually valid.  The argument must either
     be the value of the macro EOF (which has a negative value), or must be a
     non-negative value within the range representable as unsigned char.
     Passing invalid values leads to undefined behavior.

So, for signed characters, nearly half of the values represent
invalid input to the isxxxx() or toxxxx() functions, and leads to
undefined behaviour.

If you seek out the appropriate standard, I beleive SUS of some
variation, see e.g.

  http://pubs.opengroup.org/onlinepubs/7908799/xsh/tolower.html

you'll find corresponding text.  Subsequent to the cast to
"unsigned char" you'll get promotion to "int" due to the
prototype, which is fine, as long as you don't need to be able to
pass/test EOF, which I suspect isn't the case here.

Best regards,

- Håvard

fklassen:
Havard, thanks for the PR. I'll close for now, but let me know if the patch gets accepted upstream.

Issue #422, opened 2017-10-18
Unable to tcprewrite range of ports with --portmap


sdubrul:
The following doesn't seem to work in 4.2.5 on centos 7. In previous versions the portmap would allow for a range of original ports to rewrite a given port.

```
/bin/tcprewrite -r 1-100:49148 --infile=in.cap --outfile=out.cap --fixcsum
Fatal Error: Unable to parse args: From parse_args.c:tcpedit_post_args() line 181:
Unable to parse --portmap=1-100:4900
```

fklassen:
Thanks for the bug report. Scheduled for 4.2.7

fklassen:
Fixed in PR #443

PR #434, opened 2018-01-08
Change(portmap.c): tcprewrite range of ports with --portmap worked fine

nigredon1991:
I removed the unnecessary check and fixed the problem with incorrect parsing. This is my first pull request in Open Source.

### DIFF
--- a/src/tcpedit/portmap.c
+++ b/src/tcpedit/portmap.c
@@ -113,10 +113,6 @@ ports2PORT(char *ports)
             return NULL;
         }
         from_e = strtol(from_end, &badchar, 10);
-        if (strlen(badchar) != 0) {
-            free(portmap);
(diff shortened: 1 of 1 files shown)


fklassen:
Thanks for the PR, and welcome to the Open Source community. Can you describe the issue this corrects? As far as I can see, this may be introducing a memory leak.

nigredon1991:
You have issue "Unable to tcprewrite range of ports with --portmap #422". This patch fix this. In code memory free before using, therefore appeared problem. I was looking for where to get the release of memory. I think this need in src/tcprewrite.c before exit program, but I could not figure out how to do it correctly

fklassen:
Thanks. I am currently taking time from my regular job to catch up on the bug backlog. I'll start testing now.

Issue #407, opened 2017-07-01
heap-buffer-overflow in flow_decode


gy741:
Hi.

I found heap-buffer-overflow testcase.

Memory corruption occur in the ASan environment or Valgrind.

PoC Download: [PoC](https://github.com/gy741/PoC/raw/master/flow_decode)

Please confirm.

tcpreplay version: 4.2.6 (build git:v4.2.6-4-g54da347)
OS: Ubuntu 16.04.2 32bit
Command: ./tcpreplay -i lo  $FILE

Thanks.

```
Asan Log
=================================================================
==9640==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xb4a00842 at pc 0x0806b8d4 bp 0xbfb33158 sp 0xbfb33148
READ of size 2 at 0xb4a00842 thread T0
    #0 0x806b8d3 in flow_decode /home/karas/tcpreplay/src/common/flows.c:289
    #1 0x804cf18 in update_flow_stats /home/karas/tcpreplay/src/send_packets.c:371
    #2 0x804ee82 in send_packets /home/karas/tcpreplay/src/send_packets.c:575
    #3 0x805f6bc in replay_file /home/karas/tcpreplay/src/replay.c:194
    #4 0x805e6fe in tcpr_replay_index /home/karas/tcpreplay/src/replay.c:61
    #5 0x805d5ce in tcpreplay_replay /home/karas/tcpreplay/src/tcpreplay_api.c:1135
    #6 0x8055486 in main /home/karas/tcpreplay/src/tcpreplay.c:139
    #7 0xb6fba636 in __libc_start_main (/lib/i386-linux-gnu/libc.so.6+0x18636)
    #8 0x8049bf0  (/home/karas/tcpreplay/src/tcpreplay+0x8049bf0)

0xb4a00842 is located 2 bytes to the right of 32-byte region [0xb4a00820,0xb4a00840)
allocated by thread T0 here:
    #0 0xb7257144 in __interceptor_realloc (/usr/lib/i386-linux-gnu/libasan.so.2+0x97144)
    #1 0xb7196fad in grow_buffer sf-pcap.c:422

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/karas/tcpreplay/src/common/flows.c:289 flow_decode
Shadow bytes around the buggy address:
  0x369400b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[...]

Issue #410, opened 2017-07-01
heap-buffer-overflow in get_l2protocol


gy741:
Hi.

I found heap-buffer-overflow testcase.

Memory corruption occur in the ASan environment or Valgrind.

PoC Download: [PoC](https://github.com/gy741/PoC/raw/master/get_l2protocol)

Please confirm.

tcpreplay version: 4.2.6 (build git:v4.2.6-4-g54da347)
OS: Ubuntu 16.04.2 32bit
Command: ./tcpprep --auto=bridge --pcap=$FILE --cachefile=/dev/null

Thanks.

```
ASan Log
=================================================================
==21162==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xb4c02032 at pc 0x0805e545 bp 0xbfcaa528 sp 0xbfcaa518
READ of size 2 at 0xb4c02032 thread T0
    #0 0x805e544 in get_l2protocol /home/karas/gwanyeong/tcpreplay/src/common/get.c:113
    #1 0x805ea85 in get_ipv4 /home/karas/gwanyeong/tcpreplay/src/common/get.c:237
    #2 0x804d8b1 in process_raw_packets /home/karas/gwanyeong/tcpreplay/src/tcpprep.c:368
    #3 0x804ca97 in main /home/karas/gwanyeong/tcpreplay/src/tcpprep.c:146
    #4 0xb6fbb636 in __libc_start_main (/lib/i386-linux-gnu/libc.so.6+0x18636)
    #5 0x8049890  (/home/karas/gwanyeong/tcpreplay/src/tcpprep+0x8049890)

AddressSanitizer can not describe address in more detail (wild memory access suspected).
SUMMARY: AddressSanitizer: heap-buffer-overflow /home/karas/gwanyeong/tcpreplay/src/common/get.c:113 get_l2protocol
Shadow bytes around the buggy address:
  0x369803b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x369803c0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x369803d0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x369803e0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x369803f0: fa fa fa fa 00 00 00 00 00 00 00 00 00 00 00 00
[...]

Issue #409, opened 2017-07-01
heap-buffer-overflow in packet2tree


gy741:
Hi.

I found heap-buffer-overflow testcase.

Memory corruption occur in the ASan environment or Valgrind.

PoC Download: [PoC](https://github.com/gy741/PoC/raw/master/packet2tree)

Please confirm.

tcpreplay version: 4.2.6 (build git:v4.2.6-4-g54da347)
OS: Ubuntu 16.04.2 32bit
Command: ./tcpprep --auto=bridge --pcap=$FILE --cachefile=/dev/null

Thanks.

```
ASan Log
=================================================================
==21151==ERROR: AddressSanitizer: unknown-crash on address 0xb4c00bbe at pc 0x08055d50 bp 0xbff1b318 sp 0xbff1b308
READ of size 20 at 0xb4c00bbe thread T0
    #0 0x8055d4f in packet2tree /home/karas/gwanyeong/tcpreplay/src/tree.c:717
    #1 0x8055094 in add_tree_ipv4 /home/karas/gwanyeong/tcpreplay/src/tree.c:524
    #2 0x804e13d in process_raw_packets /home/karas/gwanyeong/tcpreplay/src/tcpprep.c:459
    #3 0x804ca97 in main /home/karas/gwanyeong/tcpreplay/src/tcpprep.c:146
    #4 0xb6f8a636 in __libc_start_main (/lib/i386-linux-gnu/libc.so.6+0x18636)
    #5 0x8049890  (/home/karas/gwanyeong/tcpreplay/src/tcpprep+0x8049890)

0xb4c00bd0 is located 0 bytes to the right of 32-byte region [0xb4c00bb0,0xb4c00bd0)
allocated by thread T0 here:
    #0 0xb7227144 in __interceptor_realloc (/usr/lib/i386-linux-gnu/libasan.so.2+0x97144)
    #1 0xb7166fad in grow_buffer sf-pcap.c:422

SUMMARY: AddressSanitizer: unknown-crash /home/karas/gwanyeong/tcpreplay/src/tree.c:717 packet2tree
Shadow bytes around the buggy address:
  0x36980120: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x36980130: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x36980140: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[...]

Issue #408, opened 2017-07-01
heap-buffer-overflow in get_l2len


gy741:
Hi.

I found heap-buffer-overflow testcase.

Memory corruption occur in the ASan environment or in Valgrind.

PoC Download: [PoC](https://github.com/gy741/PoC/raw/master/get_l2len)

Please confirm.

tcpreplay version: 4.2.6 (build git:v4.2.6-4-g54da347)
OS: Ubuntu 16.04.2 32bit
Command: ./tcpprep --auto=bridge --pcap=$FILE --cachefile=/dev/null

Thanks.

```
Asan Log
=================================================================
==21113==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xb4b007dc at pc 0x0805e86f bp 0xbfed6f98 sp 0xbfed6f88
READ of size 2 at 0xb4b007dc thread T0
    #0 0x805e86e in get_l2len /home/karas/gwanyeong/tcpreplay/src/common/get.c:174
    #1 0x805ea5c in get_ipv4 /home/karas/gwanyeong/tcpreplay/src/common/get.c:229
    #2 0x804d8b1 in process_raw_packets /home/karas/gwanyeong/tcpreplay/src/tcpprep.c:368
    #3 0x804ca97 in main /home/karas/gwanyeong/tcpreplay/src/tcpprep.c:146
    #4 0xb6f64636 in __libc_start_main (/lib/i386-linux-gnu/libc.so.6+0x18636)
    #5 0x8049890  (/home/karas/gwanyeong/tcpreplay/src/tcpprep+0x8049890)

0xb4b007dc is located 11 bytes to the right of 1-byte region [0xb4b007d0,0xb4b007d1)
allocated by thread T0 here:
    #0 0xb7201144 in __interceptor_realloc (/usr/lib/i386-linux-gnu/libasan.so.2+0x97144)
    #1 0xb7140fad in grow_buffer sf-pcap.c:422

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/karas/gwanyeong/tcpreplay/src/common/get.c:174 get_l2len
Shadow bytes around the buggy address:
  0x369600a0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x369600b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x369600c0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[...]

PR #435, opened 2018-01-09
fix heap-buffer-overflow in flow_decode()

GabrielGanne:
for all datalinks, check whether caplen > l2 header len
this should fix issues #407 #408 #409 #410


fklassen:
Merging to upstream test branch

PR #448, opened 2018-01-21
Enhancement #416 non blocking stdin

fklassen:


src/tcprewrite.c
```diff
@@ -149,6 +152,10 @@ main(int argc, char *argv[])
 #ifdef ENABLE_DMALLOC
     dmalloc_shutdown();
 #endif
+
+    /* avoid making STDIN non-blocking */
+    ioctl(0, FIONBIO, &nb);
```
fklassen: @dlangille some questions/comments

1. PR #416 has a compile issue fixed int this PR
2. I don't understand why this is implemented just before the exiting the program.
3. Is this fixing any particular issue?
4. If so, should it not be applied to other `main()` procedures (e.g. `tcpreplay`)?

I am pulling this PR out of 4.3 project until I receive clarification.

dlangille: I am relaying for the dev who found the issue.

tcprewrite (and likely tcpreplay see: https://stackoverflow.com/questions/16402363/read-fails-after-tcpreplay-with-error-0-resource-temporarily-unavailabl) sets STDIN as non-blocking and then doesn't set it back to blocking on exit, which causes reads to fail after exit. Example:

$ tcprewrite -C -i in.pcap -o out.pcap
$ cat > foo
cat: -: Resource temporarily unavailable

fklassen: 👍

fklassen: I applied your change to all executables. Thanks for the contribution.

dlangille: Much appreciated.  Thanks for finding that other stuff.

Issue #418, opened 2017-09-25
When replaying at captured rate, tcpreplay does not respect timing for second packet


c-f-s-s:
During some targeted testing which involved pcaps containing only a single TCP connection, I found that tcpreplay invariably generated the second packet pretty much immediately after the first, regardless of how much time difference there was between it and the first packet.

Tcpreplay seems to do a pretty good job of honoring the time deltas for the rest of the capture.

### Setup info
FreeBSD 10.3-RELEASE amd64

Ethernet Converged Network Adapter X520-SR2

tcpreplay version: 4.2.6 (build git:)
Cache file supported: 04
Compiled against libdnet: 1.12
Compiled against libpcap: 1.4.0
64 bit packet counters: enabled
Verbose printing via tcpdump: enabled
Packet editing: disabled
Fragroute engine: enabled
Default injection method: bpf send()
Optional injection method: netmap

### Reproduction info
Seems to reproduce regardless of whether or not I use the netmap drivers, and whether I use the system's 1Gb motherboard port or the Intel 10Gb nic.

Example: 
sudo tcpreplay -i ix0 --netmap <buffer file>

fklassen:
@c-f-s-s sounds fairly serious. I'm planning to catch up with bugs next week, so I'll take a look.

c-f-s-s:
Thanks Fred! I just wrote up one more that I discovered recently related to the -T flag.

_________________________________________
Cory Simonson
Engineer, Software Quality Assurance
VIAVI Solutions – viavisolutions.com
P:    952-358-3874
E:    cory.simonson@viavisolutions.com<mailto:cory.simonson@viavisolutions.com>

fklassen:
Have not been able to reproduce these issues in Linux. Will have to set up a FreeBSD environment that is using netmap. Will defer this and #419 until after 4.3 beta1

fklassen:
Fixed in PR #505

fklassen:
Fix warnings

fklassen:
Close in favour of issue #506

PR #464, opened 2018-01-28
Bug #418 2nd packet no delay

fklassen:


fklassen:
close in favour of #506

PR #468, opened 2018-03-25
resolve possible null pointer dereference

chipitsine:
issue detected by cppcheck

[src/fragroute/bget.c:814] -> [src/fragroute/bget.c:818]: (warning) Either the condition 'buf!=NULL' is redundant or there is overflow in pointer subtraction.


fklassen:
Thanks for the PR

Issue #474, opened 2018-05-15
nics unavailable when testnic specified in configure


rainey:
This is maybe the same issue initially reported in bug #178 - Based on the responses to that, it seems the intention is for the --with-testnic configure option to not affect the main programs, but it does.

Having a look through the configure script, it seems that when the --with-testnic option is specified, ENABLE_PCAP_FINDALLDEVS will become undefined, and the resulting builds will not have the --listnics option available, and for example in src/tcpreplay_api.c you have this, which causes no interfaces to be available to tcpreplay.

```
#ifdef ENABLE_PCAP_FINDALLDEVS  
   ctx->intlist = get_interface_list();  
#else  
  ctx->intlist = NULL;  
#endif
```

I'm testing this with arch linux, with pcap version 1.8.1 installed.

rainey:
Fixing this in configure.ac was fairly straightforward.  Should I submit a pull request?

fklassen:
👍

fklassen:
Merged to branch appneta:Bug_#474_enable_pcap_findalldevs_with_testnic but not merging to 4.3. Instead moving forward with fix for #178.

PR #475, opened 2018-05-16
Bug #474 enable pcap_findalldevs when --with-testnic is specified

rainey:
This setting was being skipped when --with-testnic=xxx was specified to ./configure.

Also enables osx_frameworks var when the option is specified when building for OS X.  This should make the option have only the intended effect of setting test test nic values on the build.

There may be some other issues with the current structure.  E.g. lines such as AC_DEFINE([HAVE_LINUX], [1], [Building Linux]) are never reached with testnic defined.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1712,6 +1712,22 @@ AC_ARG_WITH(testnic,
     AC_HELP_STRING([--with-testnic=NIC], [Select which network card to use for testing]),
     [ nic1=$withval
       nic2=$withval
+  
+disable_pcap_findalldevs=no
+osx_frameworks=no
+
+case $host in
+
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -1712,6 +1712,22 @@ AC_ARG_WITH(testnic,
...
+osx_frameworks=no
+
+case $host in
+
+  *-apple-darwin*)
+  
+      if test x$libpcap_version_096 = xno ; then
+        disable_pcap_findalldevs=yes
+      fi
+	  osx_frameworks=yes
```
fklassen: Thanks for the PR. Nit .. white space.

PR #482, opened 2018-08-22
Out of tree build

GabrielGanne:
Hi,

These are proposed change to allow out-of-tree build.
I've tested with the following : 
``` bash
mkdir build && cd build && ../autogen.sh && ../configure && make && make test
```

I've split that Request into two commits: the first one of which should be doing only this.
However, doing so, I've noted a lot of similarities in the plugin Makefiles (in the part I was changing).
So I'm also proposing to add a new variable for the dependency list which should factorize the code a bit.
If you decide you like that second commit, feel free to merge them both.

BR


GabrielGanne:
Please know that I've had lots of warnings when reconfiguring the sources due to my version of autogen and autotools (such as subdir-objects missing, or AC_LANG_SOURCE misconfiguration).

I've not tackled such issues here. I may try myself to them later if I have time.
Anyway be aware I might not have tested in all the right environments.

PR #495, opened 2018-10-19
Coverity scan

fklassen:


fklassen:
This PR should never merge, otherwise will continually run Coverity on every commit.

Issue #469, opened 2018-04-16
Tcpreplay resends pcap automatically in the night 


voilaviola:
Hello,
I am testing my IDS by replaying pcaps with tcprewrite and tcpreplay, and I noticed a strange thing during my test. I used tcprewrite to change the ips of pcap1.pcap to public ips, and then used tcpreplay to send it to IDS. Here's what I found in my IDS:
![image](https://user-images.githubusercontent.com/9637973/38797568-2d982724-415f-11e8-9e2c-b85429065982.png)
I replayed pcap1.pcap at around 15h, but I got 2 beaconings of same alert(same alert type, source and destination) in the night. My Wireshark and firewall records do have captured the same packets in the night, that means tcpreplay is still sending the pcap 4 hours after it stopped.
This is the only pcap that I found bizzare. The same issue doesn't occur when I replay other pcaps.
By the way, the alert triggered by this pcap is _[snort]ET INFO UPnP Discovery Search Response vulnerable UPnP device 2_
The original pcap1.pcap does not have such a time distribution, so I assume it has nothing to do with the timestamp in pcap. Is it possible that tcpreplay continues sending pcaps in the buffer after it stops?

Please find here the commands I used:
_sudo tcprewrite -s 423 --fixcsum -v -d -i pcap0.pcap -o pcap1.pcap
sudo tcpreplay -i eth1 pcap1.pcap_

Thank you in avdance,

fklassen:
What version are you running?

voilaviola:
I'm using tcpreplay and tcprewrite 3.4.4

fklassen:
That is very old. I suggest using [the latest version](https://github.com/appneta/tcpreplay/releases/latest). Let me know if you still see the failure.

voilaviola:
Hi, I updated tcpreplay and it 's solved. Please close this issue, and thanks again

fklassen:
👍

PR #511, opened 2018-10-24
Enhancement #469 POC attempt at raw sockets

fklassen:
Request is to try sending using raw sockets.

This fails because the IP stack will not accept packets that do not
include matching DLT/IP.

Here is the error I get:

reading from file -, link-type EN10MB (Ethernet)
13:41:40.1540154500 IP 172.16.230.1 > 172.16.230.200: ICMP echo reply, id 50176, seq 10, length 64
Warning in send_packets.c:send_packets() line 646:
Unable to send packet: Error with PF_PACKET send() [1]: Destination address required (errno = 89)


do11:
So the problem is if you can not send arbitrary mac address then you don't want to send packets at all?

For tests purposes, it is not always necessary to send arbitrary mac address but more important to make iptables/stack see the packets, so slightly limited send ability is still useful. It could be non-default and limitation could be stated in docs.

fklassen:
@do11 issue is not target mac/address but the source mac/address. If someone wants to hack the PR up, I will leave this branch here for future experimenting. Since this feature will not work with PCAP files generated outside of the current machine, this feature is not supportable.

do11:
But this feature should work with PCAP files generated on the current machine. I thought `tcpreplay` is a program to facilitate tests, but you are intentionally limiting its functionality. And even stating it will not be accepted (by some strange non-logical (in Vulcan way) reason).

fklassen:
@do11 I'm not saying that I will never do this feature. I just feel I am not ready to do this for 4.3.

I am concerned about supporting this option. I get flooded with the same old question as to why you cannot replay a TCP PCAP to a server and have it respond. I consider this to be a similar feature.

I could be convinced otherwise, especially if I got some assistance with testing and documenting the feature in a Wiki. If you or someone else is willing to harden this request, I may consider this for 4.3.1.

PR #535, opened 2019-01-29
ip address 

dsugisawa-mixi:
We use the tcpreplay-tcprewrite tool to generate an evaluation packet of nat system - whitebox switch.
And We would like to use actual network traffic packets as much as possible in load test packets.

So, we added the ip address range setting function to the tcprewrite tool.

By adjusting the pcap file with the tcprewrite tool mentioned above and playing it with dpdk-pktgen, it is possible to perform the high-load test necessary for the nat system.


fklassen:
Thanks for the PR. I'll review and consider submission for next major release.

fklassen:
Merging into test branch for consideration to merge into feature branch

src/tcpedit/tcpedit_opts.def
```diff
@@ -58,6 +58,49 @@ you can reuse the same seed value to recreate the traffic.
...
+    arg-type    = string;
+    max         = 1;
+    descrip     = "Rewrite IP addresses to be src/dst";
+    doc         = <<- EOText
+Takes a pair of colon delimited IPv4 addresses which will be used to rewrite
+all traffic to appear to be based on the src/dst IP addresses.
+
+IPv4 Example:
+@example
+--endpoints=198.18.50.2:198.19.0.2
```
fklassen, line 97: @dsugisawa-mixi  Is this a typo? Please correct and create PR that merges to this branch.

Also, the help has to explain what this set of parameters does. We need more than just how to use the functions. Without looking at the code, I cannot understand what they do, and which functions rely on each other.

Maybe have a look at `--pnat` as an example. Also, have a look at it to see if there is overlap.

src/tcpedit/edit_packet.c
```diff
@@ -1202,3 +1202,61 @@ is_multicast_ipv6(tcpedit_t *tcpedit, struct tcpr_in6_addr *addr)
     return 0;
 }
 
+
+/**
+ * based on  which is generated via the range_[src/dst] and range.
```
fklassen, line 1207: @dsugisawa-mixi need a better description. Also, would it make sense to add IPv6 support?

fklassen (commented):
@dsugisawa-mixi the code looks mostly acceptable, but the parameter help is a bit confusing. For now I'll target 4.5.0 (future feature release). If you can make the help more self-explanatory, I may consider for 4.4.0.

Issue #537, opened 2019-02-12
NULL pointer dereference in get_layer4_v6() 


SegfaultMasters:
**Description** - we observed a NULL pointer dereference occured in function get_layer4_v6 () located at get.c .The same be triggered by sending a crafted pcap file to the tcpreplay-edit binary. It allows an attacker to cause Denial of Service (Segmentation fault) or possibly have unspecified other impact.

**Command** - tcpreplay-edit -r 80:84 -s 20 -b -C -m 1500 -P --oneatatime -i $INTERFACE $POC 

**POC** - [REPRODUCER](https://github.com/SegfaultMasters/covering360/blob/master/tcpreplay/NP_REPL?raw=true)

**Debug** -

**GDB** -

```
Program received signal SIGSEGV, Segmentation fault.
[ Legend: Modified register | Code | Heap | Stack | String ]
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── registers ────
$rax : 0x0 
$rbx : 0x00005555557b7bb8 → 0x0000337330706e65 ("enp0s3"?)
$rcx : 0xfb 
$rdx : 0x334 
$rsp : 0x00007fffffffd8f0 → 0x000000fbffffd940
$rbp : 0x00007fffffffd920 → 0x00007fffffffd990 → 0x00007fffffffd9d0 → 0x00007fffffffda70 → 0x00007fffffffdbc0 → 0x00007fffffffdd10 → 0x00007fffffffdd40 → 0x00007fffffffddd0
$rsi : 0x123 
$rdi : 0x00005555557be236 → 0x6e0000000000cb00
$rip : 0x000055555557674a → <get_layer4_v6+222> movzx eax, BYTE PTR [rax]
$r8 : 0x2f 
$r9 : 0x00005555557be200 → 0x0000000000000000
$r10 : 0x00005555557b9700 → 0x0000000000000000
$r11 : 0x00007ffff78d6000 → <__fread_chk+0> push r13
$r12 : 0x00007ffff7bbb954 → 0x6800424d30314e45 ("EN10MB"?)
$r13 : 0x00007fffffffe2f0 → 0x000000000000000e
$r14 : 0x0 
$r15 : 0x0 
$eflags: [zero CARRY PARITY adjust SIGN trap INTERRUPT direction overflow RESUME virtualx86 identification]
[...]

Issue #536, opened 2019-02-07
 NULL pointer dereference get_ipv6_l4proto()


SegfaultMasters:
**Description** - we observed a NULL pointer dereference occured in function` get_ipv6_l4proto()` located at `get.c` .The same be triggered by sending a crafted pcap file to the tcpreplay-edit binary. It allows an attacker to cause Denial of Service (Segmentation fault) or possibly have unspecified other impact.

**Command** - tcpreplay-edit -r 80:84 -s 20 -b -C -m 1500 -P --oneatatime -i $INTERFACE $POC 

**POC** - [REPRODUCER](https://github.com/SegfaultMasters/covering360/blob/master/tcpreplay/TR_NP?raw=true)

**Debug** -

**GDB** -

```
Program received signal SIGSEGV, Segmentation fault.
[ Legend: Modified register | Code | Heap | Stack | String ]
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────[ registers ]────
$rax : 0x0 
$rbx : 0x5555557aac20 → 0x00005555557b56a0 → 0x0000337330706e65 ("enp0s3"?)
$rcx : 0xed 
$rdx : 0x8 
$rsp : 0x7fffffffdad0 → 0x000000c5ffffdb20
$rbp : 0x7fffffffdb00 → 0x00007fffffffdb70 → 0x00007fffffffdbb0 → 0x00007fffffffdc50 → 0x00007fffffffdda0 → 0x00007fffffffdef0 → 0x00007fffffffdf20 → 0x00007fffffffdfb0
$rsi : 0xc5 
$rdi : 0x7ffff7fcd6ae → 0x7ffff7b8f5600000
$rip : 0x55555557694c → <get_ipv6_l4proto+194> movzx eax, BYTE PTR [rax]
$r8 : 0x2f 
$r9 : 0x7ffff7e4d010 → 0x0000000000000000
$r10 : 0x5555557b9700 → 0x0000000000000000
$r11 : 0x7ffff78d6000 → <__fread_chk+0> push r13
$r12 : 0x5555555599e0 → <_start+0> xor ebp, ebp
$r13 : 0x7fffffffe4d0 → 0x000000000000000e
$r14 : 0x0 
$r15 : 0x0 
$eflags: [zero CARRY PARITY adjust SIGN trap INTERRUPT direction overflow RESUME virtualx86 identification]
[...]

Issue #356, opened 2017-04-13
Build fails on Ubuntu 12.04 32-bit


fklassen:
Cannot run configure. Can work around it with `./configure --host=i386-linux-gnu`

fklassen:
Fixed in release v4.2.3.

PR #544, opened 2019-03-06
stability fixes

GabrielGanne:
Hi,

Those are fix proposal for issues #536 and #537.

However, the #356 reproducer is in big endian for some reason, and when working on little endian, it fails to load because the read capture length is bigger than the maximum.
However both issues really look the same to me and the fixes are localized, so I'm confident about my proposal.
Please review.

Best regards


fklassen:
@GabrielGanne thanks for the PR. I am rebasing into a branch for test, documentations, changelog and versioning.

PR #551, opened 2019-03-20
makefile: remove some duplicated SOURCES entries

GabrielGanne:
This removes files that were listed several times from the *_SOURCES
variables.


fklassen:
thanks for the PR

PR #555, opened 2019-05-02
Warnings

GabrielGanne:
Hi,

This is to silence some warnings that appeared when I migrated from debian 9 -> 10 which comes with gcc 8.3 (over 6.3).
It addresses implicit fallthrough and unused function parameter warnings.

This does not cover them all: there are for example still warnings in makeshell.c's text_to_var() function, because gcc does not understand the NOTREACHED comments (which I find quite explicit). I did not find it a good idea to change those.

Best regards,


fklassen:
@GabrielGanne thanks for the PR. I will merge as part of next point release.

BTW, if you are interested in becoming a maintainer and contributing directly to the repo, please contact me at tcpreplay at appneta dot com

PR #561, opened 2019-06-04
Fix netmap_tx_queues_empty

ds-j-ot:


src/common/netmap.c
```diff
@@ -549,25 +549,17 @@ bool netmap_tx_queues_empty(void *p)
...
-        /* current ring is empty- go to next */
-        ++sp->cur_tx_ring;
-        if (sp->cur_tx_ring > sp->last_tx_ring) {
-            /* last ring */
-            sp->cur_tx_ring = 0;
-            return true;
-        }
 
-        txring = NETMAP_TXRING(sp->nm_if, sp->cur_tx_ring);
+    for (int i = sp->cur_tx_ring; i <= sp->last_tx_ring; i++) {
```
fklassen: Thanks for the PR. Is this change something that affects all versions of netmap? I am concerned about backwards compatibility, i.e. do we have to #ifdef or check for netmap version numbers?

ds-j-ot:
Does not fix the problem in any case

src/common/netmap.c
ds-j-ot: Sorry, as it seems this is not enough to fix the issue... I will investigate further

PR #567, opened 2019-08-13
configure.ac: fix --without-libdnet

ffontaine:
Allow the user to disable libdnet check through --without-libdnet.
This option is useful to cross-compile tcpreplay to avoid build failures
if libdnet is installed on host but not on target, see
https://bugs.buildroot.org/show_bug.cgi?id=12096

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


fklassen:
Thanks for the PR. Will apply next major release.

fklassen:
Changed from "enhancement" to "bug" and moved to 4.3.4

PR #575, opened 2020-04-24
gcc-10 warnings fixes

GabrielGanne:
Hi,

In my setup, this leaves two warning:

A `cast-function-type` warnings caused by libpcap
In my opinion, the code calling libpcap is valid, and the warning can safely be ignored.

Another saying that the <sys/sysctl.h> header is deprecated and will be removed.
I'll just leave the warning for later.

Best regards,

### DIFF
--- a/libopts/boolean.c
+++ b/libopts/boolean.c
@@ -64,8 +64,8 @@ optionBooleanVal(tOptions * opts, tOptDesc * od)
         long  val = strtol(od->optArg.argString, &pz, 0);
         if ((val != 0) || (*pz != NUL))
             break;
-        /* FALLTHROUGH */
     }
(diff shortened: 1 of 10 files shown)


fklassen:
Thanks for the PR @GabrielGanne. I will include it in the next fix release, coming shortly.

fklassen:
> A `cast-function-type` warnings caused by libpcap
> In my opinion, the code calling libpcap is valid, and the warning can safely be ignored.

I got a tip from Debian guys to try `-fcommon` flag to get rid of issues, but it undesirable to do so. I have been attempting to spin up a gcc-10 sandbox to attempt to recreate.

As usual, thanks for the help.

Issue #578, opened 2020-05-19
[Bug] tcpreplay-edit ——heap-use-after-free in get_ipv6_next() at get.c:454


14isnot40:
**Describe the bug**
A heap-use-after-free bug was discovered in tcpreplay-edit binary, during the structure 'exthdr' points to the memory address released after use. The issue is being triggered in the function get_ipv6_next() at common/get.c:454.

**To Reproduce**
1. Compile tcpreplay according to the default configuration
```bash
./configure CFLAGS="-g -O0 -fsanitize=address"
```
2. execute command
```bash
tcpreplay-edit -r 80:84 -s 20 -b -C -m 1500 -P --oneatatime -i lo $poc
```
[poc](https://github.com/14isnot40/vul_discovery/blob/master/tcpreplay/tcpreplay-edit/pocs/huaf_get_ipv6_next) can be found here.

**Expected behavior**
An attacker can exploit this vulnerability by submitting a malicious pcap that exploits this issue. This will result in a Denial of Service (DoS), potentially Information Exposure and even Code Execution when the application attempts to process the file.

**Screenshots**
ASAN Reports
```bash
==64632==ERROR: AddressSanitizer: heap-use-after-free on address 0x611000009c87 at pc 0x0000004412bc bp 0x7fffffffd500 sp 0x7fffffffd4f0
READ of size 4 at 0x611000009c87 thread T0
    #0 0x4412bb in get_ipv6_next /home/test/Desktop/evaulation/tcpreplay/src/common/get.c:454
    #1 0x44112b in get_layer4_v6 /home/test/Desktop/evaulation/tcpreplay/src/common/get.c:409
    #2 0x428410 in do_checksum /home/test/Desktop/evaulation/tcpreplay/src/tcpedit/checksum.c:66
    #3 0x42033a in fix_ipv4_checksums /home/test/Desktop/evaulation/tcpreplay/src/tcpedit/edit_packet.c:82
    #4 0x41c8f9 in tcpedit_packet /home/test/Desktop/evaulation/tcpreplay/src/tcpedit/tcpedit.c:354
    #5 0x40963b in send_packets /home/test/Desktop/evaulation/tcpreplay/src/send_packets.c:552
    #6 0x418e9a in replay_file /home/test/Desktop/evaulation/tcpreplay/src/replay.c:182
    #7 0x417e73 in tcpr_replay_index /home/test/Desktop/evaulation/tcpreplay/src/replay.c:59
[...]

PR #587, opened 2020-06-01
Bug #578 get_ipv6_next use after free

fklassen:


fklassen:
After fix ...

```
Warning in sendpacket.c:sendpacket_open_pf() line 943:
Unsupported physical layer type 0x0304 on lo.  Maybe it works, maybe it won't.  See tickets #123/318
Warning in replay.c:replay_file() line 138:
../tcpreplay-pcaps/huaf_get_ipv6_next was captured using a snaplen of 255 bytes.  This may mean you have truncated packets.

Fatal Error in send_packets.c:send_packets() line 553:
 Error editing packet #1: From edit_packet.c:fix_ipv4_checksums() line 73:
Invalid packet: Expected IPv4 packet: got 6
```

Issue #566, opened 2019-08-07
[Bug] tcpreplay stop to generate unique-ip with --unique-ip option after some cycles


Polexy:
**Describe the bug**
I have pcap with only one udp packet. 
I playing it like:

> tcpreplay -i ens7 -p 30000 -K -l 0 --unique-ip my.pcap

after ~ 11 minutes / 16 million cycles i see tcpreplay start to generate 2 same packets, after 38 million packets tcpreplay start to generate 4 same packets and etc.

> 12:58:15.045217 IP 7.190.152.209.1947 > 12.67.16.142.51802: UDP, length 22
> 12:58:15.045250 IP 7.190.152.209.1947 > 12.67.16.142.51802: UDP, length 22
> 12:58:15.045284 IP 7.190.152.208.1947 > 12.67.16.143.51802: UDP, length 22
> 12:58:15.045316 IP 7.190.152.208.1947 > 12.67.16.143.51802: UDP, length 22

i take pcap with 2 different packets and tried to play it on same speed(30k pps) and i see this situation start after ~38 million packets ~22min.
so, this bug reproduced after ~ 16-17 million cycles.

**To Reproduce**
1. Find/Create pcap with one udp packet
2. Play this pcap. (For example tcpreplay -i ens7 -p 30000 -K -l 0 --unique-ip my.pcap)
3. After ~16-17 million cyclyes on speed 30000 pps ~10-12min - look to packets(i used tcpdump)

**Expected behavior**
I thought tcpreplay can generate ~2 milliard unique packets. ( (255^4)/2 -because every cycle: dst.ip +1, src.ip -1)

**System :**
Ubuntu 16.04.6 LTS
tcpreplay version: 4.3.1 (build git:v4.3.1)
Copyright 2013-2018 by Fred Klassen <tcpreplay at appneta dot com> - AppNeta
Copyright 2000-2012 by Aaron Turner <aturner at synfin dot net>
The entire Tcpreplay Suite is licensed under the GPLv3
Cache file supported: 04
Not compiled with libdnet.
Compiled against libpcap: 1.7.4
64 bit packet counters: enabled
Verbose printing via tcpdump: enabled
Packet editing: disabled
Fragroute engine: disabled
Default injection method: PF_PACKET send()
Optional injection method: netmap

fklassen:
Thanks for the detailed report. It was very helpful in reproducing the issue. Fixed in PR #600

PR #600, opened 2020-06-03
Bug #566 Fix float to integer rounding issue

fklassen:
At very high iteration counts '--unique-ip' calculation
(which support floating point e.g. 0.75) has accuracy
issues. Corrected by converting to integer earlier,
and multiplying numerator/denominator by 1000.

### DIFF
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -89,7 +89,7 @@ static uint32_t get_user_count(tcpreplay_t *ctx, sendpacket_t *sp, COUNTER count
  */
 static void
 fast_edit_packet_dl(struct pcap_pkthdr *pkthdr, u_char **pktdata,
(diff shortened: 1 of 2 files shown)


src/send_packets.c
```diff
@@ -460,7 +460,8 @@ static void increment_iteration(tcpreplay_t *ctx)
     if (options->unique_ip) {
         assert(options->unique_loops > 0.0);
         ctx->unique_iteration =
-                (COUNTER)((float)ctx->iteration / options->unique_loops) + 1;
+                ((ctx->iteration * 1000) / (COUNTER)(options->unique_loops * 1000.0))
+                + 1;
     }
```
fklassen, line 465: This is the actual fix. The reset of the changes are to prevent wrapping for logging.

Issue #611, opened 2020-06-25
[Bug] tcprewrite fails with "From tcpedit.c:tcpedit_packet() line 192: Packet length 98 is to short to contain an IPv6 header for DLT 0x0001"


yp-isi:
**Describe the bug**
TCP rewrite fails to parse input pcap file, exits with the error:
```
Fatal Error: Error rewriting packets: From tcpedit.c:tcpedit_packet() line 192:
Packet length 98 is to short to contain an IPv6 header for DLT 0x0001
```

**To Reproduce**
1. Run `tcprewrite` as follows:
```
tcprewrite -vv -i afail1.pcap -o /dev/null --pnat="9.9.0.0/16:1.1.0.0/16"
```
2. Observe the output:
```
reading from file -, link-type EN10MB (Ethernet)
17:11:17.1592871077 IP6 4950:3aa3:964b:a356:60a3:7dea:a836:3200 > 4efa:695:ee19:12d9:4906:6f8b:8bbb:6bd6: frag (0|36) 63927 > domain: 0 [0q] (28)

Fatal Error: Error rewriting packets: From tcpedit.c:tcpedit_packet() line 192:
Packet length 98 is to short to contain an IPv6 header for DLT 0x0001
```

**Expected behavior**
Not to fail.  Even if the packet is not parse-able, it should either (a) drop it, (b) overwrite IPs, but not checksum, (c) leave such packets as is.

**Screenshots**
N/A

**System (please complete the following information):**
 - OS: linux
 - OS fedora 31
 - Tcpreplay Version
```
tcpreplay version: 4.3.2 (build git:v4.3.2)
Copyright 2013-2018 by Fred Klassen <tcpreplay at appneta dot com> - AppNeta
Copyright 2000-2012 by Aaron Turner <aturner at synfin dot net>
The entire Tcpreplay Suite is licensed under the GPLv3
Cache file supported: 04
Compiled against libdnet: 1.12
Compiled against libpcap: 1.9.1
64 bit packet counters: enabled
Verbose printing via tcpdump: enabled
Packet editing: disabled
Fragroute engine: enabled
Injection method: PF_PACKET send()
Not compiled with netmap
```

**Additional context**
Failed pcap attached.
[afail1.pcap.gz](https://github.com/appneta/tcpreplay/files/4834199/afail1.pcap.gz)

fklassen:
[...]

PR #613, opened 2020-07-13
tcprewrite: handle malformed and unsupported packets as soft errors

GabrielGanne:
Hi,

This is in response to #611

The ticket issue is caused by the packet being rewritten to be
unsupported (fragmented IPv6 packet). The issue would be the same
for malformed packets however.

The current behavior is to fail quite brutally by stopping all processing
right away. I propose to just skip those packets instead.

Best regards,


fklassen:
> I'm not certain I am comfortable with accepting this change. It usually means that the packet capture is corrupt. I have had lots of CVE's opened because Tcpreplay was processing invalid packets and accessing uninitialized memory. This change may open the door for more bug reports.

I had a closer look, and it seems worth the risk. This is great to have, enabling edits on short caplen captures. I suspect that we may see some bugs come out of this at some point, but for now I cannot find any.

Thanks again @GabrielGanne for a great contribution.

PR #627, opened 2020-12-15
fixed simple typo, to short -> too short

ddeangelis:
There is a small typo in src/tcpedit/tcpedit.c

`to short to contain` should read `too short to contain`


fklassen (approved):
LGTM

Issue #630, opened 2021-02-19
[Bug] Packets slowly drift further and further behind when they should be sent


davecraig:
Using a simple tcpreplay e.g.

   sudo src/tcpreplay -K test.pcap

If I record another PCAP of the playout, the packets in the playout happen later and later compared to the original test.pcap. The reason is that the timing is all done using deltas between packets and errors in timing accumulate. This has a huge impact on playback of time sensitive PCAPs e.g. UDP video containing PCR timestamps.

I have a suggested patch for this which switches to comparing times since the first packet was sent.

fklassen:
I have not seen this before. I need more context. Please fill in the bug report.

**Describe the bug**
A clear and concise description of what the bug is.

**To Reproduce**
Steps to reproduce the behavior:
1. 
2. 
3. 
4. 

**Expected behavior**
A clear and concise description of what you expected to happen.

**Screenshots**
If applicable, add screenshots to help explain your problem.

**System (please complete the following information):**
 - OS: [e.g. macOS]
 - OS version
 - Tcpreplay Version [e.g. 4.3.2]

**Additional context**
Add any other context about the problem here.

davecraig:
tcpreplay slowly loses time meaning that packets get more and more delayed as the playout continues. An easy way to reproduce this is:

1. Create a PCAP e.g. `randpkt -b 1400 -t udp -c 100000 test.pcap`
2. Playout the pcap with tcpreplay: `sudo tcpreplay -K --multiplier=10000 test.pcap`
3. Note how long it takes (Actual: 100000 packets (71786397 bytes) sent in 10.10 seconds in my case) and then double check tcpdump the data to see how accurately the packets were sent out.

`randpkt` creates 1 packet each second, so the playout is done at 10000 speed which gives around ~56Mbps. In my test the tcpdump looked like this:

```
No.	Time	Source	Destination	Protocol	Length	Info	Absolute time
26	*REF*	208.21.2.184	10.1.1.99	UDP	1055	9587 → 206 [BAD UDP LENGTH 33828 > IP PAYLOAD LENGTH] Len=33820	26
27	0.000009	208.21.2.184	10.1.1.99	UDP	795	4171 → 62106 [BAD UDP LENGTH 59293 > IP PAYLOAD LENGTH] Len=59285	27
[...]

PR #631, opened 2021-02-19
send_packet: Avoid clock drift by using time since first packet

davecraig:
Prior to this change, time between packets was calculated as deltas
and over time errors would quickly add up. The work here switches
to use the wall time since the first packet was sent as the reference
which is compared to the PCAP packet timestamp relative to the first
packet. In this way there are no errors and clock drift is avoided.
As a result tcpreplay can now be used to play back very time sensitive
captures such as video streams with PCR.

This is a suggested fix for https://github.com/appneta/tcpreplay/issues/630

### DIFF
--- a/src/common/utils.h
+++ b/src/common/utils.h
@@ -36,6 +36,8 @@ typedef struct {
     struct timeval end_time;
     struct timeval pkt_ts_delta;
     struct timeval last_print;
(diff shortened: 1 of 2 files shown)


fklassen:
@davecraig this may have introduced Bug #674

PR #648, opened 2021-03-16
clean up string manipulation in error exit

guijan:
Hello, I think it's self evident this is a better way to manipulate strings.

A quick grep for 'EBUF' in the source tree says there could be other places that could be improved on, I could send more patches if so.


fklassen:
> Hello, I think it's self evident this is a better way to manipulate strings.
> 
> A quick grep for 'EBUF' in the source tree says there could be other places that could be improved on, I could send more patches if so.

Yes, more PR's like this are welcome. They may or may not make it for 4.3.4 release (coming soon).

Feel free to add your name to docs/CREDITS, or if you give me your name I'll add it.

PR #656, opened 2021-04-09
Update strlcpy.c and strlcat.c

guijan:
The other BSD functions in the /lib directory are outdated too but I'm not familiar with them and I haven't checked if there are any breaking changes to their interfaces.

Additionally, have you considered the [libbsd](https://libbsd.freedesktop.org/wiki/) library? It's a popular library so it's on pretty much every distro repo and it runs on Linux, Mac, and it doesn't run on Windows but there Cygwin provides BSD functions instead and Microsoft's WSL is just Ubuntu as far as we're concerned. 

The library provides all the BSD extensions in the /lib directory of tcpreplay's source tree and many more, it also has an "overlay" mode that introduces the functions in the same standard headers that the BSDs extend, and if the libc it was compiled for implements the functions it doesn't include them. Cygwin also provides the BSD functions in the standard headers, so portability is only a matter of depending on the library on Linux, WSL, and Mac.

### DIFF
--- a/lib/strlcat.c
+++ b/lib/strlcat.c
@@ -1,7 +1,7 @@
-/*	$OpenBSD: strlcat.c,v 1.11 2003/06/17 21:56:24 millert Exp $	*/
+/*	$OpenBSD: strlcat.c,v 1.19 2019/01/25 00:19:25 millert Exp $	*/
 
 /*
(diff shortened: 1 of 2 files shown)


fklassen (approved):

fklassen:
Thanks for the PRs. Will consider `libbsd` as a future enhancement.

PR #682, opened 2021-09-06
Directories of pcaps as arguments

halver94:
Sometimes your pcaps are buried deep into some directories and sub-directories. Depending on how many you want to pass to tcpreplay and how long their name are, you may be limited by the shell command line length.
Of course you can do another directory with some symbolic links, but it can be more convenient that tcpreplay loads directly a directory or a list of directories.


fklassen:
@halver94 thanks for the PR. Staging to a testing branch, targeting 4.4.0 release

PR #755, opened 2022-10-20
Simulate packet loss

dsseng:
Useful when testing debugging UDP-based software like realtime audio streaming, simulates non-ideal network conditions.

### DIFF
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -592,8 +592,9 @@ send_packets(tcpreplay_t *ctx, pcap_t *pcap, int idx)
             send_len -= l2len;
         }
 
-        /* write packet out on network */
-        if (sendpacket(sp, send_data, send_len, &pkthdr) < (int)send_len) {
+        /* write packet out on network, skipping it in random cases when asked for drop */
+        if ((options->drop == 1.0f || rand() > options->drop * RAND_MAX) &&
+            sendpacket(sp, send_data, send_len, &pkthdr) < (int)send_len) {
             warnx("Unable to send packet: %s", sendpacket_geterr(sp));
             continue;
         }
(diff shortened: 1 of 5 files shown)


fklassen:
This is an option that should be in `tcpreplay-edit` not `tcpreplay`. There are other issues that this PR would introduce that would make me reluctant to accept this. This includes breaking dual-interface sends, `tcpprep` etc. Also, there are no tests included.

I'll shelve this for now, but may reconsider if more work is done.

sh7dm:
Yes, it would make more sense in tcpreplay-edit. Basically the work here was done for a simple test and then sent for upstreaming.

fklassen:
PR's are always welcome. I'll have a look at it when I work on an enhancement release (4.5).

sh7dm:
Great, thank you for your effort!

fklassen (changes requested):
Thanks for the PR — packet loss simulation is a good addition for testing against non-ideal network conditions. Tracking this feature in #1055; a few requested changes before merging:

1. **Drop the RNG seeding.** `srand(time(NULL))` in `tcpreplay.c`'s `main()` isn't needed — no need to seed the random number generator here.

2. **Use a 0-100 scale, and call it "percent loss."** Right now `--drop` takes a 0.0-1.0 fraction. Please switch to an integer/float 0-100 percentage instead (e.g. `--loss=25` drops ~25% of packets), and rename the option/field from "drop" to reflect that it's percent loss, not a fraction.

3. **Update the `--help`/man page text** for the renamed option, and make sure it explicitly states that the loss is **random** — i.e. packets are dropped probabilistically per-packet, not on a fixed pattern (e.g. not "every 4th packet").

One more thing worth double-checking while you're in there: `options->drop == 1.0f` is currently used as the sentinel for "no `--drop` given" (skips the random check entirely, always sends). With a fraction-based scale, `1.0` would naturally mean "drop everything," so a user explicitly passing that value gets the opposite of what they'd expect — the "not set" sentinel and a real "drop 100%" value collide. Worth using a separate flag/sentinel (e.g. `HAVE_OPT(DROP)` at the call site, or a `bool` alongside the percentage) so an explicit 100% loss actually works.

PR #796, opened 2023-05-02
Nanosec accurate packet processing

plangarbalint:
The aim of this change is to make packet processing possible with nanosecond accuracy and to handle sub sleep times by only nanosleep()'ing. 

- change struct timeval to struct timespec where it was necessary 
- to have backward compatibility with gettimeofday, get_current_time wrapper function is used to retrieve current time (either using gettimeofday or clock_gettime based on the value of _POSIX_C_SOURCE)
- pcap_open_offline is change to pcap_open_offline_with_tstamp_precision in replay_file function, as a result nanosec accuracy is not lost now directly when pcap file is read
- extend nanosleep_sleep function with clock_nanosleep feature using TIMER_ABSTIME in order to avoid relative sleep when using nanosleep()
- extend loop delay feature with nanosec delaying in case --timer=nano is set, earlier only microsec delay was possible

### DIFF
--- a/src/common/timer.c
+++ b/src/common/timer.c
@@ -91,13 +91,13 @@ init_timestamp(struct timespec *timestamp)
     timesclear(timestamp);
 }
 
-int get_time_of_day(struct timespec *ts) {
-    struct timeval tv;
-    int success = gettimeofday(&tv, NULL);
-    TIMEVAL_TO_TIMESPEC(&tv, ts);
-    return success;
-}
-
-int clock_get_time(struct timespec *ts){
-    return clock_gettime(CLOCK_REALTIME, ts);
+int get_current_time(struct timespec *ts){
+    #ifdef _POSIX_C_SOURCE >= 199309L
+        return clock_gettime(CLOCK_REALTIME, ts);
+    #else
+        struct timeval tv;
+        int success = gettimeofday(&tv, NULL);
+        TIMEVAL_TO_TIMESPEC(&tv, ts);
+        return success;
+    #endif
 }
--- a/src/common/timer.c
+++ b/src/common/timer.c
@@ -92,7 +92,7 @@ init_timestamp(struct timespec *timestamp)
 }
 
 int get_current_time(struct timespec *ts){
-    #ifdef _POSIX_C_SOURCE >= 199309L
+    #if defined _POSIX_C_SOURCE  && _POSIX_C_SOURCE >= 199309L
         return clock_gettime(CLOCK_REALTIME, ts);
     #else
         struct timeval tv;
--- a/src/common/timer.c
+++ b/src/common/timer.c
@@ -86,8 +86,18 @@ void timesdiv(struct timespec *tvs, COUNTER div)
 }
 
 void
-init_timestamp(timestamp_t *ctx)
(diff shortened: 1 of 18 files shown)


src/common/timer.c
```diff
@@ -86,8 +86,18 @@ void timesdiv(struct timespec *tvs, COUNTER div)
...
 }
 
+int get_current_time(struct timespec *ts){
+    #if defined _POSIX_C_SOURCE  && _POSIX_C_SOURCE >= 199309L
+        return clock_gettime(CLOCK_REALTIME, ts);
+    #else
+        struct timeval tv;
+        int success = gettimeofday(&tv, NULL);
+        TIMEVAL_TO_TIMESPEC(&tv, ts);
+        return success;
```
fklassen, line 101: nice that this supports older OS's. How much testing has there been on this?

fklassen (commented):
Thanks for the PR. Will target for next major release.

src/common/timer.c
plangarbalint: A comprehensive test on this is definitely missing, I made some user tests with basic setups. However there are many include guards, with if, elseif etc... branches, they all need to be tested separately. Just to mention one, it wasn't tested how it works with netmap

src/common/timer.c
```diff
@@ -86,8 +86,18 @@ void timesdiv(struct timespec *tvs, COUNTER div)
...
-init_timestamp(timestamp_t *ctx)
+init_timestamp(struct timespec *timestamp)
 {
-    timerclear(ctx);
+    timesclear(timestamp);
 }
 
+int get_current_time(struct timespec *ts){
+    #if defined _POSIX_C_SOURCE  && _POSIX_C_SOURCE >= 199309L
+        return clock_gettime(CLOCK_REALTIME, ts);
```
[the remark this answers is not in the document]
SPYFF: Since the PR also using absolute time semantics for more precise sleeping, both `CLOCK_MONOTONIC`  and `CLOCK_REALTIME` would work. However AFAIK POSIX only defines `CLOCK_REALTIME` mandatory. Probably every sane OS implements both clocksource, but I think realtime chosen because of portability concerns.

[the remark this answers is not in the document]
fklassen: I'll put it through it's paces when I start work on versin 4.5. I'll see if I can get my confidence up that it won't affect IoT and oldert stuff.

src/send_packets.c
```diff
@@ -1013,32 +1019,32 @@ static void calc_sleep_time(tcpreplay_t *ctx, struct timeval *pkt_ts_delta,
...
 
             /*
-             * bits * 1000000 divided by bps = microseconds
+             * bits * 1000000000 divided by bps = nanosecond
              *
              * ensure there is no overflow in cases where bits_sent is very high
              */
             if (bits_sent > COUNTER_OVERFLOW_RISK && bps > 500000)
-                next_tx_us = (bits_sent * 1000) / bps * 1000;
+                next_tx_ns = (bits_sent * 1000) / bps * 1000000;
```
fklassen, line 1035: This will take some amount of testing. Some overflows take > 1 week to occur.

PR #814, opened 2023-07-21
Fix UDP checksum recalculation

green-nsk:
```
If the computed  checksum  is zero,  it is transmitted  as all ones (the
equivalent  in one's complement  arithmetic).   An all zero  transmitted
checksum  value means that the transmitter  generated  no checksum  (for
debugging or for higher level protocols that don't care).
```
https://www.ietf.org/rfc/rfc768.txt


fklassen:
Closing as invalid. As per the above spec, transmitting a checksum of 0 is valid (at least for IPv4) in cases where the protocol does not need a checksum. The only time you would do this is the COMPUTED checksum is zero.

green-nsk:
I assume the whole point of computing the checksum it to transmit non-null value. Otherwise you could as well `return 0;`, it would also be technically valid

Issue #831, opened 2023-11-29
[Bug] tcprewrite fails to change addresses for ipv6 fragment packets


ChuckCottrill:
**Describe the bug**

TCP rewrite fails to parse input pcap file, exits with a `TCPEDIT_SOFT_ERROR` due to detection of `TCPR_IPV6_NH_FRAGMENT` in `src/common/get.c`.

**Expected behavior:**
TCP rewrite should change source and destination addresses (and all other packet header revisions),
even when the `TCPR_IPV6_NH_FRAGMENT` proto is set.

**To Reproduce**
1. Run tcprewrite and observe the output, as follows
```
tcprewrite -vv -i afail1.pcap -o afail1out.pcap -c cc/afail1.cache -e [2600:bead::1]:[2600:cafe::1] ;
reading from file -, link-type EN10MB (Ethernet), snapshot length 65535
19:11:17.1592871077 IP6 4950:3aa3:964b:a356:60a3:7dea:a836:3200 > 4efa:695:ee19:12d9:4906:6f8b:8bbb:6bd6: frag (0|36) 63927 > 53: 0 [0q] (28)
```
2. run tcpdump to show output unchanged:
```
tcpdump -r afail1-out.pcap -A -XX -s 0 -c 2
reading from file afail1-out.pcap, link-type EN10MB (Ethernet), snapshot length 65535
19:11:17.562141 IP6 4950:3aa3:964b:a356:60a3:7dea:a836:3200 > 4efa:695:ee19:12d9:4906:6f8b:8bbb:6bd6: frag (0|36) 63927 > domain: 0 [0q] (28)
	0x0000:  3480 0d44 4f81 f4a7 39e3 70d4 86dd 6000  4..DO...9.p...`.
	0x0010:  0000 002c 2c37 4950 3aa3 964b a356 60a3  ...,,7IP:..K.V`.
	0x0020:  7dea a836 3200 4efa 0695 ee19 12d9 4906  }..62.N.......I.
	0x0030:  6f8b 8bbb 6bd6 1100 0000 ea26 fa23 f9b7  o...k......&.#..
	0x0040:  0035 0024 f2d8 0000 0000 0000 0000 0000  .5.$............
	0x0050:  0000 0000 0000 0000 0000 0000 0000 0000  ................
	0x0060:  0000                                     ..
```

**Expected behavior**
1. tcprewrite should not fail, and should rewrite the address:
```
[...]

Issue #488, opened 2018-10-15
Heap overflow in csum_replace4()


SegfaultMasters:
A heap-based buffer overflow was discovered in tcpreplay-edit binary, during the incremental checksum operation. The issue is being triggered in the function `csum_replace4()` at `incremental_checksum.h`, invoked by `ipv4_l34_csum_replace()` in `edit_packet.c`. 

### **Tested version:**
4.3

### **Command:**
tcpreplay-edit --portmap=80:8000 --seed=10  --cachefile=example.cache --intf1=eno1 --intf2=eno3 --decode=some --preload-pcap --verbose $POC

### **Debugging:**

```
     95  static inline void csum_replace4(__sum16 *sum, __be32 from, __be32 to)
     96  {
                // sum=0xbfffe8f0 -> [...] -> 0xa8c0b0af
->   97       *sum = csum_fold(csum_add(csum_sub(~csum_unfold(*sum), from), to)); //Buffer overflow
     98  }
```
	 

```
gef> p (unsigned short *)sum
$20 = (unsigned short *) 0xb6001742
gef> x 0xb6001742
0xb6001742:     391171182

```

### **ASAN Report** 

```
==7737==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xb6001742 at pc 0x0806a35c bp 0xbff60578 sp 0xbff60568
READ of size 2 at 0xb6001742 thread T0
    #0 0x806a35b in csum_replace4 /home/loginsoft/ACE/tcpreplay/src/tcpedit/incremental_checksum.h:97
    #1 0x806acf1 in ipv4_l34_csum_replace /home/loginsoft/ACE/tcpreplay/src/tcpedit/edit_packet.c:181
    #2 0x806afdb in ipv4_addr_csum_replace /home/loginsoft/ACE/tcpreplay/src/tcpedit/edit_packet.c:251
    #3 0x806b633 in randomize_ipv4 /home/loginsoft/ACE/tcpreplay/src/tcpedit/edit_packet.c:343
    #4 0x806661b in tcpedit_packet /home/loginsoft/ACE/tcpreplay/src/tcpedit/tcpedit.c:272
    #5 0x805158a in send_packets /home/loginsoft/ACE/tcpreplay/src/send_packets.c:554
    #6 0x8063193 in replay_file /home/loginsoft/ACE/tcpreplay/src/replay.c:188
[...]

PR #832, opened 2023-11-29
Handle IPv6 fragment extension header

ChuckCottrill:
**Description:**
Handle IPv6 fragment extension header
only return NULL when reached end of packet (no data)

**Explanation:**
When issue #488 was fixed (pr #496) (version=4.3.0, and carried forward to version=4.4.0), the fix prevented IPv6
fragment extension header handling. The need was to prevent addressing headers beyond packet length (to avoid
heap corruption). However, the fix prevents any further processing of the packet after fragment extension header found.
When the extension header `proto=TCPR_IPV6_NH_FRAGMENT`, we can skip that header (fixed length=32 bits),
process any further extension headers, and return a pointer to packet data.

When issue #611 was fixed (pr #613), the solution was to leave any IPv6 packets with fragment extension header
untouched (and generate `TCPEDIT_SOFT_ERROR`). This change allowed tcprewrite to continue, and these packets
could be skipped (use option `--skip-soft-errors`). This was to avoid abort/failure of the code to run on pcap having
any fragment headers.

Both of these issues are handled by this PR.
- resolves issue #496 by avoiding addressing headers beyond packet length
- resolves issue #611 by enabling tcprewrite to continue after encountering fragment packets.

**Reference:**
- see: [tcprewrite fails to change addresses for ipv6 fragment packets](https://github.com/appneta/tcpreplay/issues/831)
- see: [tcprewrite fails on IPv6 fragment extension header](https://github.com/appneta/tcpreplay/issues/611)
- see: [Heap overflow in csum_replace4](https://github.com/appneta/tcpreplay/issues/488)

### DIFF
--- a/src/common/get.c
+++ b/src/common/get.c
@@ -599,7 +599,7 @@ get_layer4_v6(const ipv6_hdr_t *ip6_hdr, const u_char *end_ptr)
             // next points to l4 data
             dbgx(3, "Go deeper due to fragment extension header 0x%02X", proto);
             exthdr = get_ipv6_next(next, end_ptr);
-            if ((NULL == exthdr) || (exthdr > end_ptr)) {
+            if ((NULL == exthdr) || ((u_char *)exthdr > end_ptr)) {
                 next = NULL;
                 done = true;
                 break;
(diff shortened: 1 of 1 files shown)


fklassen:
Thanks for the PR. Opened issue #837 to track.

Merging into a working branch for testing and documentation.

Targetting 4.5 release.

PR #853, opened 2024-05-17
add -w output.pcap command line option to direct the output to a pcap

jasonlue:
tcpreplay is a great tool to replay pcap files as a traffic generator. tcpreplay-edit is even better with the feature of --unique-ip to simulate real traffic, being fed only a small amount of seed traffic as input. However, to test a system's stability, performance and load balancing, and for troubleshooting/debugging in such an environment, we need large amount of randomly distributed, but repeated traffic as single pcap file.  

Right now there's no easy way for it.  The typical way to do it is to use tcpreplay-edit --unique-ip or t-rex  on one (virtual) machine to generate random traffic to a specific port, and then use tcpdump -i -w on another to sniff the traffic and store it to a pcap file. This isn't the most convenient way and its time consuming. Also this settings have a side effect where the played traffic causes the overall network system to respond to them, thus recorded traffic isn't pure randomized version of the seed traffic, but it includes quite a few responding traffic.

By giving tcpreplay-edit a -w <output.pcap> option, we can simply run tcpreplay-edit --unique-ip -w http_many.pcap  http.pcap to generate such a traffic.

This PR is an effort to cover a use case just like that.


fklassen:
Thanks for the PR. I'll merge into a staging branch to test. I'll try to get this into 4.5.0.

src/common/sendpacket.c
```diff
@@ -671,6 +686,9 @@ sendpacket_get_hwaddr(sendpacket_t *sp)
 
     if (sp->handle_type == SP_TYPE_KHIAL) {
         addr = sendpacket_get_hwaddr_khial(sp);
+    } else if( sp->handle_type == SP_TYPE_LIBPCAP_DUMP) {
+        sendpacket_seterr(sp, "Error: sendpacket_get_hwaddr() not yet supported for pcap dump");
```
fklassen, line 740: This will need to get implemented.

jasonlue: I think  sendpacket_get_hwaddr() only makes sense for real interfaces.  Pcap file as an output doesn't support sendpacket_get_hwaddr.

fklassen: OK, I still have this in staging and will investigate. I'll probably just change the error message.

fklassen:
Thanks for the PR. I pushed into 4.5.0. I now see the value of this option, and is in the top 3 favourite enhancements in 4.5.0.

Issue #813, opened 2023-07-19
[Bug] Double free in tcpedit_dlt_cleanup in tcprewrite


iskindar:
**Describe the bug**
tcprewrite in tcpreplay latest commit : [43693c4](https://github.com/appneta/tcpreplay/commit/43693c4a92dda227a5134ae5e71a7247ae345323),  v4.4.4 and v.4.4.3 has a double free in function  tcpedit_dlt_cleanup in plugins/dlt_plugins.c. 

**To Reproduce**

1. Get the Tcpreplay source code and build it with ASAN.
```
# Build with ASAN
export CC=gcc export CXX=g++
export CFLAGS="-g -fsanitize=address" export CXXFLAGS="-g -fsanitize=address"
./autogen.sh
./configure && make -j
```
2. Run tcprewrite with provided poc
```
tcprewrite -i poc -o /dev/null
```

The poc is available at [poc.zip](https://github.com/appneta/tcpreplay/files/12090738/poc.zip)

Please unzip it first and then feed it into the tcprewrite binary.

**Expected behavior**

The ASAN report 
```
==72056==ERROR: AddressSanitizer: attempting double-free on 0x603000000040 in thread T0:
    #0 0x7f10d91ff40f in __interceptor_free ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:122
    #1 0x555929e48229 in our_safe_free /benchmark/tcpreplay/src/common/utils.c:113
    #2 0x555929e2dbe6 in tcpedit_dlt_cleanup plugins/dlt_plugins.c:466
    #3 0x555929e1dc35 in tcpedit_close /benchmark/tcpreplay/src/tcpedit/tcpedit.c:555
    #4 0x555929e19cd1 in main /benchmark/tcpreplay/src/tcprewrite.c:146
    #5 0x7f10d8e8f082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082)
    #6 0x555929e1714d in _start (/validate_binary/tcprewrite+0x1c14d)

0x603000000040 is located 0 bytes inside of 20-byte region [0x603000000040,0x603000000054)
freed by thread T0 here:
    #0 0x7f10d91ff40f in __interceptor_free ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:122
    #1 0x555929e48229 in our_safe_free /benchmark/tcpreplay/src/common/utils.c:113
[...]

Issue #850, opened 2024-03-10
[Bug] Double-free in tcpreplay's tcprewrite utility


msxfXF:
You are opening a _bug report_ against the Tcpreplay project: we use
GitHub Issues for tracking bug reports and feature requests.

If you have a question about how to use Tcpreplay, you are at the wrong
site. You can ask a question on the [tcpreplay-users mailing list](https://sourceforge.net/p/tcpreplay/mailman/tcpreplay-users/)
or [on Stack Overflow with  [tcpreplay] tag](http://stackoverflow.com/questions/tagged/tcpreplay). 
General help is available [here](http://tcpreplay.appneta.com/).

If you have a build issue, consider downloading the [latest release](https://github.com/appneta/tcpreplay/releases/latest)

Otherwise, to report a bug, please fill out the reproduction steps
(below) and delete these introductory paragraphs.  Thanks!

**Describe the bug**
A double-free vulnerability exists within the tcprewrite utility of the tcpreplay suite. When handling specific packet capture files, tcprewrite may attempt to free the same memory location twice leading to potential code execution, denial of service, or memory corruption scenarios.

The issue occurs in the tcpedit_dlt_cleanup function, as part of the dlt_plugins.c code, and can be triggered under certain conditions, as evidenced by the provided crash file and stack trace pointing to a problem when cleaning up resources.

**To Reproduce**
1. export CC=clang && export CFLAGS="-fsanitize=address -g"
2. ./autogen.sh && ./configure --disable-shared --disable-local-libopts && make clean && make -j8
3. tcprewrite -o /dev/null -i POC.pcap

**Screenshots**
```
Warning: ./id:000000,sig:06,sync:master_write,src:000138 was captured using a snaplen of 9999 bytes.  This may mean you have truncated packets.
=================================================================
==1515869==ERROR: AddressSanitizer: attempting double-free on 0x503000000280 in thread T0:
    #0 0x55ddaa37b476 in free (/usr/local/bin/tcprewrite+0xdf476)
    #1 0x55ddaa3d0b51 in tcpedit_dlt_cleanup /home/shf/固件代码/tcpreplay/src/tcpedit/plugins/dlt_plugins.c:466:9
    #2 0x55ddaa3be500 in tcpedit_close /home/shf/固件代码/tcpreplay/src/tcpedit/tcpedit.c:555:9
    #3 0x55ddaa3ba07c in main /home/shf/固件代码/tcpreplay/src/tcprewrite.c:146:5
[...]

PR #855, opened 2024-05-19
Fix recursive tcpedit cleanup

GabrielGanne:
Assume a single tcpedit struct and return the previously allocated context if called twice.

This fixes an issue with the Juniper Encapsulated Ethernet DLT plugin which has an exception in the way the plugins works with regard to the extra buffer in question: tcpreplay works with the assumption that there only ever is a single link layer plugin which is mostly true except here: Juniper has a special call to tcpedit_dlt_copy_decoder_state() which causes the ctx and subctx to share a reference to the decoded_extra buffer, and a double free.

Fixes: #813 #850

### DIFF
--- a/src/tcpedit/plugins/dlt_plugins.c
+++ b/src/tcpedit/plugins/dlt_plugins.c
@@ -94,6 +94,12 @@ const char *tcpeditdlt_bit_info[] = {"Missing required Layer 3 protocol.",
  * Public functions
  ********************************************************************/
 
+/*
+ * Ensure init/cleanup are called only once
+ * Assume a single tcpedit struct and return the previously allocated context.
+ */
+static int tcpedit_dlt_is_initialized = 0;
+
 /**
  * initialize our plugin library.  Pass the DLT of the source pcap handle.
  * Actions:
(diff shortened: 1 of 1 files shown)


GabrielGanne:
Hi,
`cpp-linter` is failing with the following error:
```
src/tcpedit/plugins/dlt_plugins.c:21:10 [clang-diagnostic-error]: src/tcpedit/plugins/dlt_plugins.c#L21
'config.h' file not found
```
 Is it possible this is some configuration issue ? The normal build works fine.
 BR

fklassen:
> Hi, `cpp-linter` is failing with the following error:
> 
> ```
> src/tcpedit/plugins/dlt_plugins.c:21:10 [clang-diagnostic-error]: src/tcpedit/plugins/dlt_plugins.c#L21
> 'config.h' file not found
> ```
> 
> Is it possible this is some configuration issue ? The normal build works fine. BR

In 4.5.0 I made it so that the linter never fails, but it will still put inline suggestions in the PR. If the code quality goes up, I can restore the CI failure.

fklassen:
Thanks for the PR. Will merge to 4.5.0 now and then test before release.

fklassen:
I had to re-open #813 and back out this change because this PR, when applied after fixes #711 and #851, introduced a memory leak.

fklassen:
Not a problem. I much prefer getting help and becoming a tester than hunting down problems on my own. Your contributions are always welcome.

Issue #844, opened 2024-01-24
[Bug] tcprewrite provides incorrect checksum for certain ipv4 packets


ChuckCottrill:
The tcprewrite program provides incorrect checksum and modifies packet length in an undesireable manner.

**Describe the bug**
TCP rewrite produces an incorrect IP and TCP checksum for certain pcap files.
TCP rewrite appears to change packet length incorrectly, and thus produces an invalid checksum;
certain downstream processing may treat said incorrect checksum as a spoofing attempt and discard packet.

**Expected behavior:**
TCP rewrite should only change packet length when that behavior is specifically desired (command line option?).
TCP rewrite should correctly calculate IP and TCP checksum (incorrect because length incorrect).

**To Reproduce**

1. uncompress packet captures:
```
mkdir -p pcaps
unzip tcprewrite-pcaps.zip
cp tcprewrite-pcaps/pcap-original-packet-3.pcap pcaps/.
```
2. Run tcprewrite version 4.4.0 and observe the output, as follows
```
# version 4.4.0
VERSION="4.4.0"
# prepare
tcpreplay-4.4.0/src/tcpprep \
    --cidr=0.0.0.0/0 \
    --pcap=pcaps/pcap-original-packet-3.pcap \
    --cachefile=pcaps/pcap.cache
# use tcprewrite to rewrite packet addresses
tcpreplay-4.4.0/src/tcprewrite \
    --cachefile=pcaps/pcap.cache \
    --infile=pcaps/pcap-original-packet-3.pcap \
    --outfile=pcaps/cap-4.4.0-packet-out.pcap \
    --endpoints=10.200.1.1:10.200.1.2
```
3. Run tcprewrite version 4.4.1 and observe the output, as follows
```
# version 4.4.1
VERSION="4.4.1"
# prepare
tcpreplay-4.4.1/src/tcpprep \
    --cidr=0.0.0.0/0 \
    --pcap=pcaps/pcap-original-packet-3.pcap \
    --cachefile=pcaps/pcap.cache
# use tcprewrite to rewrite packet addresses
tcpreplay-4.4.1/src/tcprewrite \
    --cachefile=pcaps/pcap.cache \
    --infile=pcaps/pcap-original-packet-3.pcap \
    --outfile=pcaps/cap-4.4.1-packet-out.pcap \
    --endpoints=10.200.1.1:10.200.1.2
```
4. compare files, should be identical
```
[...]

PR #862, opened 2024-06-02
Bug #844 PR# 846: fix regression on nanosecond timestamp

fklassen:
libpcap nanosec values are reported in tv_usec.

Not so for legacy libpcap which has usec values in tv_usec.


fklassen:
Closing in favour of PR #865

PR #871, opened 2024-06-05
Feature #853 direct traffic to pcap

fklassen:


fklassen:
I'm not quite ready to move this into 4.5.0. I am still struggling to see what value this has vs. tcprewrite. I don't want to shelve this yet, but I'll review for the flowing feature release.

jasonlue:
The use scenario:

Suppose I have a seed pcap http.pcap with only a few packets. The goal is to generate a large http_many.pcap file. http_many.pcap file repeats http.pcap, but with different IP addresses.

Solution #1:
(1) write a bash to multiply http.pcap to http.more.of.the.same.pcap with wireshark's mergecap
(2) tcprewrite http.more.of.the.same.pcap, randomize the ip address, and output as http_many.pcap.

Solution #2:
tcpreplay --unique-ip  -l1000 -w http.many.pcap http.pcap

Solution #2 is siimple and straight-forward. It doesn't require another program (mergepcap) and bash file. (Or a simple app to do the same).  We only deploy tcpreplay and enjoy the benefits.

fklassen:
OK, I never thought of that scenario. I am wondering if it may be better to add `-l` and `--unique-ip` to `tcp-rewrite`, or maybe we can come up with a different solution. What are your thoughts?

Currently there is a clear distinction between `tcpreplay` and `tcprewrite`. The former writes to networks, latter writes to files. I am inclined to keep it that way. Adding `--unique-ip` to tcpreplay caused some confusion, and some argued it should have only been in tcprewrite. But it was something I added because it suited a project we were working on.

jasonlue:
adding -l --unique-ip looks like a good idea.

fklassen:
I am reconsidering this in light of feature #884, tcpreplay is producing results that are had to do with the combination `tcpprep` and `tcprewrite`. Since `-w` is taken, maybe the `-o` option can be used.

Issue #984, opened 2025-09-17
[Bug] Tcpreplay try to send the same packet continuously after it failed when use -2 parameter


sjd1105:
**Describe the bug**
Tcpreplay try to send the same packet continuously after it failed when use -2 parameter.
Maybe caused by the behavior changed in #751 from version 4.4.3.

**To Reproduce**
1. download pkt_too_long.pcap mentioned in #751 , it contains 3 packets in which the 2nd packet is too long and cannot be sent
2. use tcpreplay 4.4.3 and replay single pcap, it works well

> tcpreplay -i ens8f0 -p 1 --no-flow-stats pkt_too_long.pcap
> Warning: Unable to send packet: Error with PF_PACKET send() [2]: Message too long (errno = 90)
> Actual: 2 packets (196 bytes) sent in 2.00 seconds
> Rated: 97.9 Bps, 0.000 Mbps, 0.99 pps
> Statistics for network device: ens8f0
>         Successful packets:        2
>         Failed packets:            1
>         Truncated packets:         0
>         Retried packets (ENOBUFS): 0
>         Retried packets (EAGAIN):  0

3. replay double pcap, cause this issue
> tcpreplay -2 -i ens8f0 -I ens8f1 -p 1 --no-flow-stats pkt_too_long.pcap pkt_too_long.pcap
> Warning: Unable to send packet: Error with PF_PACKET send() [2]: Message too long (errno = 90)
> Warning: Unable to send packet: Error with PF_PACKET send() [3]: Message too long (errno = 90)
> Warning: Unable to send packet: Error with PF_PACKET send() [4]: Message too long (errno = 90)
> Warning: Unable to send packet: Error with PF_PACKET send() [5]: Message too long (errno = 90)
> Warning: Unable to send packet: Error with PF_PACKET send() [6]: Message too long (errno = 90)
> Warning: Unable to send packet: Error with PF_PACKET send() [7]: Message too long (errno = 90)
> Warning: Unable to send packet: Error with PF_PACKET send() [8]: Message too long (errno = 90)
> Warning: Unable to send packet: Error with PF_PACKET send() [9]: Message too long (errno = 90)
> Warning: Unable to send packet: Error with PF_PACKET send() [10]: Message too long (errno = 90)
> ^C User interrupt...
> sendpacket_abort
> Warning: Unable to send packet: Error with PF_PACKET send() [10]: Message too long (errno = 90)
> Actual: 2 packets (210 bytes) sent in 10.44 seconds
[...]

PR #986, opened 2026-01-19
Refs #984: sendpacket: avoid endless retries on ENOBUFS/EAGAIN

d3156:
Refs #984.

### Problem
When using the PF_PACKET injection method, `sendpacket()` may enter an endless retry loop if `send()` keeps returning `ENOBUFS` or `EAGAIN` continuously (e.g., with very large frames or under sustained buffer pressure). This results in 100% CPU usage and tcpreplay never progresses to the next packet.

### Change
Introduce a bounded retry loop for transient `ENOBUFS`/`EAGAIN` failures:
- Add a small sleep on retries to avoid busy-looping and unnecessary kernel load.
- Stop retrying after `max_retry_count` attempts and return an error so tcpreplay can continue with the next packet.

### How to test
1. Build tcpreplay from this branch.
2. Replay a pcap that triggers persistent `ENOBUFS`/`EAGAIN` on PF_PACKET.
3. Verify that tcpreplay no longer hangs in an infinite retry loop and instead proceeds (or reports failure) after the retry limit.

Values are currently hard-coded and can be made configurable if maintainers prefer.

### DIFF
--- a/src/common/sendpacket.c
+++ b/src/common/sendpacket.c
@@ -257,6 +257,8 @@ sendpacket(sendpacket_t *sp, const u_char *data, size_t len, struct pcap_pkthdr 
                                   * prevent page misses on stack
                                   */
     static const size_t buffer_payload_size = sizeof(buffer) + sizeof(struct pcap_pkthdr);
+    size_t retry_count = 0;
+    const size_t max_retry_count = 100;
 
     assert(sp);
 #ifndef HAVE_LIBXDP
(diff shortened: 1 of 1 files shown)


fklassen:
Thanks for this — the fix is merged, folded into #996 with a couple of review fixups on top:

- `usleep(0)` doesn't actually sleep on any target libc, so it wasn't preventing busy-looping as intended; replaced with a real `usleep(100)` backoff, matching the existing convention in `src/common/txring.c`.
- Added an explicit `sendpacket_seterr()` call when the retry cap is hit, so callers get an accurate error instead of a stale/empty buffer.
- Minor style cleanup (label indentation, trailing whitespace).

Merged in 426706a on `Bug_#984_endless_retries_on_ENOBUFS_EAGAIN`, and #984 is closed. Closing this one as superseded — credited in `docs/CREDIT` and `docs/CHANGELOG`. Thanks again for the fix!

Issue #895, opened 2024-07-15
[Bug] Build-Depends on GNU autogen


fklassen:
**Describe the bug**

As reported in Debian Bug#1076243: tcpreplay: Build-Depends on GNU autogen

Source: tcpreplay
Version: 4.4.4-1
Severity: normal
User: [ametzler@debian.org](mailto:ametzler@debian.org)
Usertags: autogen-eol

Good morning,

tcpreplay build-depends on GNU autogen, which has reached EOL and IMHO
is unlikely to be revived.

The best replacement for autogen's AutoOpts part I know (but I do not
know everything) is the toolchain gnutls upstream built for their
conversion:
1. One-time onversion of .def to json with
https://gitlab.com/dueno/parse-autogen and some handholding.
2. Build-time generation from json input of option-parser and docs with
https://gitlab.com/gnutls/cligen (python).

cu Andreas

fklassen:
Merged PR #991 to v4.5.3-beta1. Many thanks to  @GabrielGanne for an excellent fix.

fklassen:
Reopening due to issues found in PR #911.

Reverting this PR for now, and delaying for future feature release 4.6.

Issues that need to be reviewed:

- breaks `sudo make test`
- no `--load-opts` option (may choose to drop this)
- option `--more-help` fails, e.g. 
```
❯ src/tcpreplay --more-help
Warning in /Users/fklassen/git/tcpreplay/src/common/interface.c:get_interface_list() line 80:
May need to run as root to get access to all network interfaces.
tcpreplay: unrecognized option `--more-help'
tcpreplay - Replay network traffic stored in pcap files
Usage:  tcpreplay [ -<flag> [<val>] | --<name>[{=| }<val>] ]... <pcap_file(s)>

  -d, --dbug=num           Enable debugging output
  -q, --quiet              Quiet mode
  -T, --timer=str          Select packet timing mode: select, ioport, gtod, nano
      --maxsleep=num       Sleep for no more then X milliseconds between packets
  -v, --verbose            Print decoded packets via tcpdump to STDOUT
  -A, --decode=str         Arguments passed to tcpdump decoder
  -K, --preload-pcap       Preloads packets into RAM before sending
  -c, --cachefile=str      Split traffic via a tcpprep cache file
  -2, --dualfile           Replay two files at a time from a network tap
  -i, --intf1=str          Client to server/RX/primary traffic output interface
  -I, --intf2=str          Server to client/TX/secondary traffic output interface
[...]

PR #991, opened 2026-06-08
migrate autoopts -> getopt + asciidoctor

GabrielGanne:
## Standards checklist:

- [x ] The PR title is descriptive.
- [x] The PR doesn't replicate another PR which is already open.
- [x] The code is mine or it's from somewhere with an MIT-compatible license.
- [x] The code is efficient, to the best of my ability, and does not waste computer resources.
- [x] The code is stable and I have tested it myself, to the best of my abilities.
- [x] The code passes `sudo make test`

## Changes:

Hi, this is a proposal to address issue https://github.com/appneta/tcpreplay/issues/895

I have chosen not to follow the way recommended in the issue description (mostly because of personal preferences). This is proposing to remove libopt and autoopt entirely and to use asciidoctor for documentation generation.
This means I'm breaking the current idea of having one true source for both documentation and code. Although I find that idea is intellectually pleasing, it looks to me this is not the way anymore, at least I was not able to find a good replacement that would fit the current need.

I took the liberty of changing some of the interfaces like some removal of `--short-help` and `--long-help` in favor of simple `--help`, and some of the man page section have had their lines moved (eg. short help moved from synopsys to description). I think it's better that way, but I have no strong opinion and can restore them.

## Other comments:

I have used AI to help generate the `*_args.c` and `*_args.h` files.

A note on the license: I have chosen to re-write some macros from autoopts to reduce the amount of changes. libopts was LGPLv3/modified-BSD (GPL-compatible) and tcpreplay is GPLv3 so this should be fine.

A note on the choice to add a dependency on asciidoctor. The ruby part is annoying but this is only a problem for the maintainers, the source distribution builds just like before. Also the project looks to be well established (git, wireshark, redhat's openshift ...) and is packaged by all major distributions.

Best regards


fklassen:
thanks @GabrielGanne for an excellent PR

fklassen:
Unfortunately had to revert this PR in #1000 

Need to cherry-pick a new PR for next feature release and fix issues documented in #895

PR #1002, opened 2026-07-17
sendpacket: stop inhibiting libbpf's default XDP program load for AF_XDP TX

### LINKED ISSUE #956: [Bug] Cannot run with `--xdp`: Packet sending exited with error
**Describe the bug**

When adding `--xdp`, no packet is sent and `Packet sending exited with error` occurs

**To Reproduce**
1. Download 4.5.2-beta1, and build following the doc
2. `sudo tcpreplay --xdp -i eth1 some-pkts.pcap
3. See the error

**Expected behavior**
Tcpreplay works like without `--xdp`, sending packets.

**Screenshots**

Image

 - OS: Ubuntu 24.04 in vagrant-libvirt's vm (kvm qemu)
 - OS version
   - 24.04.2 LTS (Noble Numbat)
   - 6.8.0-64-generic
 - Tcpreplay Version: 4.5.2-beta1
 - NIC: Intel Corporation Ethernet Controller XL710 for 40GbE QSFP+ (rev 02)

```console
❯ sudo tcpreplay -V
tcpreplay version: 4.5.2-beta1 (build git:v4.5.2-beta1) (debug)
Copyright 2013-2024 by Fred Klassen <tcpreplay at appneta dot com> - AppNeta
Copyright 2000-2012 by Aaron Turner <aturner at synfin dot net>
The entire Tcpreplay Suite is licensed under the GPLv3
Cache file supported: 04
Not compiled with libdnet.
Compiled against libpcap: 1.10.4
64 bit packet counters: enabled
Verbose printing via tcpdump: enabled
Packet editing: disabled
Fragroute engine: disabled
Injection method: PF_PACKET send()
Not compiled with netmap
Optional injection method: AF_XDP
```

**Additional context**
This seems to be a duplicated of #898, but with different NIC. But no error message seen in `kern.log`.

```console
❯ tail /var/log/kern.log
2025-07-17T09:55:44.974345+00:00 vm-sec-4 kernel: ipmi_si: Unable to find any System Interface(s)
2025-07-17T09:55:45.062374+00:00 vm-sec-4 kernel: Bluetooth: Core ver 2.22
2025-07-17T09:55:45.062400+00:00 vm-sec-4 kernel: NET: Registered PF_BLUETOOTH protocol family
2025-07-17T09:55:45.062403+00:00 vm-sec-4 kernel: Bluetooth: HCI device and connection manager initialized
2025-07-17T09:55:45.062405+00:00 vm-sec-4 kernel: Bluetooth: HCI socket layer initialized
2025-07-17T09:55:45.062407+00:00 vm-sec-4 kernel: Bluetooth: L2CAP socket layer initialized
2025-07-17T09:55:45.062408+00:00 vm-sec-4 kernel: Bluetooth: SCO socket layer initialized
2025-07-17T09:55:45.720678+00:00 vm-sec-4 kernel: virtio_net virtio2 eth0: entered promiscuous mode
2025-07-17T10:57:42.115553+00:00 vm-sec-4 kernel: i40e 0000:03:00.0 eth1: entered promiscuous mode
2025-07-17T10:57:42.128777+00:00 vm-sec-4 kernel: i40e 0000:03:00.0 eth1: left promiscuous mode
```

I wonder where can I ger more debug info for this error.

fklassen:
## Summary

Fixes #956 — `--xdp` failing with `EINVAL` ("XDP is either not supported by this underlying network driver, or it has a bug") on real NICs: Intel XL710 (i40e), Intel X520 (ixgbe), and virtio_net, per multiple reporters in the issue thread.

**Root cause**: `sendpacket_open_xsk()` → `create_xsk_socket()` set `XSK_LIBBPF_FLAGS__INHIBIT_PROG_LOAD`, skipping libbpf's default auto-attached XDP program (tcpreplay is TX-only and doesn't need the RX-redirect behavior that program provides). But several drivers only enable their `ndo_xdp_xmit` TX datapath once *some* native XDP program is attached to the interface — without one, `xsk_socket__create()` still succeeds, but the first zero-copy send fails with `EINVAL`.

**Fix**: drop the inhibit flag so libbpf auto-loads its default program. That's enough to trigger the driver's XDP TX setup, even though tcpreplay doesn't use the RX side of it. `xsk_socket__delete()` (already called on close) tears the auto-loaded program down.

This is the same class of workaround `xdp-tools`' `xdp-trafficgen` uses for the same driver behavior, but simpler — uses libbpf's own default program instead of shipping/attaching a custom `xdp_pass` eBPF object.

## Second bug found along the way

While testing this, hit a pre-existing, unrelated bug: `./configure`'s `LIBXDP` feature probe rejected every real libxdp install. It locally redefined `struct xsk_socket` with guessed fields (leftover from the old header-only sample `xsk.h`, predating this project linking the real libxdp shared library) and passed `&xsk` — a `struct xsk_socket *` — to `xsk_socket__create()`, whose first param is `struct xsk_socket **`. Pointer-indirection mismatch → probe always failed to compile → `HAVE_LIBXDP` never got defined → `--xdp` silently compiled out regardless of what was installed. Fixed in the same PR since the #956 fix can't be built or tested without it.

## Verification

Confirmed by @fklassen:
- Compiles cleanly with the fixed `configure` probe (`LIBXDP for AF_XDP socket: yes`)
- `--xdp` works on i40e (XL710) and ixgbe (X520) — the two drivers reported broken in #956
- No regression on veth / virtio_net (previously-working paths)

## Test plan

- [x] Compiles cleanly (CI has `libxdp-dev`)
- [x] Real-hardware verified on i40e (XL710) and ixgbe (X520)
- [x] No regression on veth / virtio_net

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1476,14 +1476,7 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
 #include <xdp/xsk.h>
 #include <sys/socket.h>
 ]], [[
-    struct xsk_socket {
-        struct xsk_ring_cons *rx;
-        struct xsk_ring_prod *tx;
-        struct xsk_ctx *ctx;
-        struct xsk_socket_config config;
-        int fd;
-    };
-    struct xsk_socket xsk;
+    struct xsk_socket *xsk = NULL;
 	struct xsk_ring_cons *rxr = NULL;
 	struct xsk_ring_prod *txr = NULL;
     int queue_id = 0;
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,8 +1,9 @@
 Unreleased
-    - sendpacket: --xdp AF_XDP zero-copy TX failing with EINVAL on drivers (i40e, ixgbe,
-      virtio_net, ...) that require a native XDP program attached before ndo_xdp_xmit works;
-      let libbpf auto-load its default program instead of inhibiting it (#956) [UNVERIFIED on
-      affected hardware - needs testing before promoting past beta]
+    - sendpacket: fix --xdp AF_XDP zero-copy TX failing with EINVAL on drivers (i40e, ixgbe)
+      that require a native XDP program attached before ndo_xdp_xmit works; let libbpf
+      auto-load its default program instead of inhibiting it (#956, #1002)
+    - configure: fix LIBXDP feature probe rejecting every real libxdp install due to a
+      pointer-indirection mismatch in the xsk_socket__create() compile check (#1002)
     - fragroute: fix build failure on macOS 13/14 from TAILQ_FOREACH_REVERSE argument-order
       collision between bundled lib/queue.h and system <sys/queue.h> (#981, #1001)
     - tcpprep: fix buffer overflow in check_dst_port() reading TCP/UDP dest port on truncated
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -1,4 +1,8 @@
 Unreleased
+    - sendpacket: --xdp AF_XDP zero-copy TX failing with EINVAL on drivers (i40e, ixgbe,
+      virtio_net, ...) that require a native XDP program attached before ndo_xdp_xmit works;
+      let libbpf auto-load its default program instead of inhibiting it (#956) [UNVERIFIED on
+      affected hardware - needs testing before promoting past beta]
     - fragroute: fix build failure on macOS 13/14 from TAILQ_FOREACH_REVERSE argument-order
       collision between bundled lib/queue.h and system <sys/queue.h> (#981, #1001)
     - tcpprep: fix buffer overflow in check_dst_port() reading TCP/UDP dest port on truncated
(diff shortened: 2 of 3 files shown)


fklassen:
Found and fixed a second, bigger bug while you were testing this: the `./configure` probe for LIBXDP support was broken independent of #956.

`AC_COMPILE_IFELSE` locally redefined `struct xsk_socket` with guessed fields (leftover from the old header-only sample xsk.h, predating this project linking the real libxdp shared library) and passed `&xsk` — a `struct xsk_socket *` — to `xsk_socket__create()`, whose first param is `struct xsk_socket **`. Pointer-indirection mismatch, so the probe fails to compile against any current libxdp, and `checking for LIBXDP XDP packet sending support` always says `no` even with `-lxdp` found and `xdp/libxdp.h` present — exactly your symptom. `HAVE_LIBXDP` never gets defined, so `--xdp` silently compiles out no matter what's installed.

Fixed in configure.ac (latest commit on this branch) — declares `struct xsk_socket *xsk = NULL` and passes `&xsk` instead, matching the real signature (which is also what src/common/sendpacket.c already does correctly).

Can't fully verify `checking ... yes` myself — this dev box is macOS, no libxdp at all. Could you re-pull this branch, re-run `./autogen.sh && ../configure --with-testnic=enp0s5 --with-testnic2=enp0s5 | grep -i xdp` and confirm it now reports `yes`? That'll also be the first real compile+test of the #956 driver-attach fix itself.

PR #1012, opened 2026-07-17
tcpr_random: fix signed left-shift undefined behavior (does not fix #1011 - see comments)

### LINKED ISSUE #1011: [Bug] macOS arm64 (Apple Silicon): 3 tcprewrite tests produce different output than macOS Intel / Linux arm64 - 2 real cross-platform bugs found
## Summary

\`sudo make test\` on v4.5.3-beta1 passes on macOS Intel and Linux arm64, but 3 tests fail on macOS arm64 (Apple Silicon): \`rewrite_portmap\`, \`rewrite_range_portmap\`, \`rewrite_sequence\`.

**Environment where it fails:**
\`\`\`
Darwin P16GRYP44Q 25.5.0 Darwin Kernel Version 25.5.0: Tue Jun  9 22:28:24 PDT 2026; root:xnu-12377.121.10~1/RELEASE_ARM64_T6020 arm64
\`\`\`

Passes on: macOS Intel (same v4.5.3-beta1 code), and Linux arm64 (same code).

Reporter supplied the actual output produced on macOS arm64 for the 3 failing tests (as \`test2.*\` files) so the golden files could be diffed against what that platform actually produces. Analysis below is from comparing those against the current checked-in goldens (\`test/test.rewrite_*\`).

**This is not one bug - it's (at least) two distinct, real, cross-platform-portability bugs**, both found by byte-level + semantic diffing of the pcap outputs (not just "arm64 is weird" - actual root causes in tcpedit's C code). Neither is a golden-file/test-harness artifact: I first ruled out pcap container endianness (the global header's byte-order-tagging differs between the files but is semantically equivalent and correctly auto-detected by any conformant reader - confirmed by parsing both files endian-aware and comparing decoded packet content, not raw bytes).

---

## 1. `rewrite_sequence` - CONFIRMED root cause: signed left-shift UB in `tcpr_random()`

Test command: `tcprewrite -i test.pcap -o test.rewrite_sequence --tcp-sequence 42`

`--tcp-sequence 42` derives a single 32-bit `tcp_sequence_adjust` value from the seed (42), then adds it to every packet's TCP seq/ack (`src/tcpedit/rewrite_sequence.c`, `rewrite_seqs()`). Since the adjustment is a single value applied uniformly, I checked whether the *offset* between the two platforms' output was constant across all differing packets - it is: **exactly 2165354059 on every single one of the 116 differing packets**, which means this isn't per-packet drift, it's one seed-derivation computing a different `tcp_sequence_adjust` on this platform.

Traced it to `tcpr_random()` in `src/common/utils.c:436-458`, called 5 times seeded with 42 (`src/tcpedit/parse_args.c:169-175`):

```c
uint32_t
tcpr_random(uint32_t *seed)
{
    unsigned int next = *seed;
    int result;

    next *= 1103515245;
    next += 12345;
    result = (int)(next / 65536) % 2048;

    next *= 1103515245;
    next += 12345;
    result <<= 10;
    result ^= (int)(next / 65536) % 1024;

    next *= 1103515245;
    next += 12345;
    result <<= 10;                                   // <-- HERE
    result ^= (int)(next / 65536) % 1024;

    *seed = next;
    return result;
}
```

`result` is a **signed** `int`. After the first two stages, `result` can hold up to 21 significant bits (`(2047 << 10) | 1023` = 2097151). The second `result <<= 10` then shifts that up to ~31 bits - close enough to the sign bit boundary that, depending on the exact bit pattern, **a set bit can be shifted into or past the sign bit of a signed int**. Left-shifting a signed integer such that the result isn't representable in the type is undefined behavior in C (C11 6.5.7p4). Different compiler backends are free to handle that UB differently - and evidently macOS arm64's clang codegen (targeting Apple Silicon) does something different from macOS Intel's and Linux arm64's, despite nominally "the same" toolchain family.

The doc comment on this function literally says *"Implementation of rand_r that is consistent across all platforms"* - the entire point of not using libc's `rand()`/`rand_r()` was to avoid exactly this class of cross-platform inconsistency, but the signed-shift UB reintroduces it.

**Fix direction** (not yet implemented, wanted to confirm root cause first): change `result` to `unsigned int`/`uint32_t`. Unsigned left-shift overflow is well-defined (mod 2^32 wraparound) in C, which eliminates the UB and would make the function's platform-consistency guarantee actually hold.

---

## 2. `rewrite_range_portmap` - symptom fully characterized, root cause NOT yet pinned down

Test command: `tcprewrite -i test.pcap -o test.rewrite_range_portmap -r 1-100:49148` (maps port range 1-100 to port 49148).

Confirmed via full-packet decode (not just byte diffs) that **the range-portmap rewrite has zero effect at all on macOS arm64**:

- Repo golden (Intel Mac / Linux arm64): **97 of 179 packets** contain port 49148 (the mapped target) - portmap is working.
- macOS arm64 actual output: **0 of 179 packets** contain port 49148 anywhere - portmap did nothing; every port that should have been remapped (e.g. packet 12: dport 80, which is in range 1-100, stays 80 instead of becoming 49148) is left completely untouched.

This is not a partial/data-dependent failure (which would point at something like a matching-logic edge case) - it's total: as if the `-r` option were silently ignored on that platform.

**What I checked and ruled out**, to save time for whoever picks this up next:
- **Not a pcap-container/endianness artifact** - confirmed via endian-aware parsing that packet *content* genuinely differs, not just header byte-order tagging.
- **Not the "extra uninitialized linked-list node" pattern I initially suspected** - `ports2PORT()`'s range-parsing loop in `src/tcpedit/portmap.c` (lines ~118-125) does allocate one extra trailing node beyond what's needed for an N-port range (a real, if currently harmless, off-by-one worth fixing separately), but `safe_malloc()` zeroes memory (`src/common/utils.c` `our_safe_malloc()`), so that extra node's `from`/`to` are `0`, not garbage - it can never spuriously match a real port (1-65535), and every legitimately-mapped port in the range still gets its own correctly-populated node regardless of platform. This doesn't explain an all-or-nothing failure.
- **Not a bitfield-ABI issue** - checked whether `tcp_hdr_t`'s port fields or `ipv4_hdr_t`'s version/IHL bitfields could be laid out differently on Apple's arm64 ABI (a real, known category of macOS-arm64-specific bug). `th_sport`/`th_dport` are plain `uint16_t`, not bitfields, so this doesn't apply to the port fields directly.

**Where I'd look next** (couldn't go further without the actual arm64 hardware to run a debugger/verbose build against): the "all or nothing" pattern points toward the portmap *option not being applied at all* rather than a numeric edge case in the matching logic (`map_port()` in `portmap.c`) - i.e. either `parse_portmap()` isn't populating `tcpedit->portmap` correctly on this platform, or the `HAVE_OPT(PORTMAP)` check / AutoOpts-generated option parsing itself behaves differently there. Worth noting AutoOpts/libopts (this project's CLI option-parsing library, see #895) is old, unmaintained upstream, and has at least one other confirmed real bug already (#931, unrelated) - it predates Apple Silicon entirely and its arm64-macOS behavior may simply never have been exercised much. Running with `-d 5` (verbose) or a debugger on the actual arm64 machine to check whether `tcpedit->portmap` is populated at all after option parsing would be the fastest way to confirm or rule this out.

---

## 3. `rewrite_portmap` (simple, non-range: `-r 80:8080`) - blocked on missing data

The supplied `test2.rewrite_portmap` file is **0 bytes** - can't analyze it yet. If this test is also failing, it'd be useful evidence either way: if the *simple* (non-range) portmap code path (`ports2PORT()`'s final `else` branch in `portmap.c`, which doesn't have the loop/off-by-one pattern the range branch does) also fails, that argues more strongly for the "option isn't being applied at all" theory from #2 above rather than anything specific to the range-parsing branch. If it passes, that narrows things back toward the range-specific code path.

**Requesting a valid (non-empty) `test2.rewrite_portmap` to complete this analysis.**

---

## Reproduction

```sh
# on macOS arm64
cd build && sudo make test   # rewrite_portmap, rewrite_range_portmap, rewrite_sequence fail

# to compare outputs directly:
tcprewrite -i test/test.pcap -o /tmp/out.rewrite_sequence --tcp-sequence 42
tcprewrite -i test/test.pcap -o /tmp/out.rewrite_range_portmap -r 1-100:49148
diff test/test.rewrite_sequence /tmp/out.rewrite_sequence
diff test/test.rewrite_range_portmap /tmp/out.rewrite_range_portmap
```

🤖 Diagnosis assisted by Claude Code

fklassen:
## Summary

Fixes a real signed left-shift undefined-behavior bug in `tcpr_random()`. **Does NOT fix #1011** — see the correction below and the actual fix in #1013.

`tcpr_random()` (`src/common/utils.c`) is a custom `rand_r()`-alike, explicitly documented as *"consistent across all platforms"* — the entire point of not using libc's `rand()`/`rand_r()`. It declared its accumulator as **signed** `int`:

```c
int result;
...
result <<= 10;   // x2
```

After the first two mixing stages, `result` can carry up to ~21 significant bits. The second `result <<= 10` shifts that up to ~31 bits — close enough to the sign bit that a set bit can be shifted into or past it. Left-shifting a signed `int` such that the result isn't representable in the type is **undefined behavior** in C (C11 6.5.7p4). This is real UB and worth fixing regardless of the outcome below.

## Fix

Declare `result` as `unsigned int` instead. Unsigned left-shift overflow is well-defined (mod 2^32 wraparound) in C, which removes the UB entirely. All five call sites already treat the return value as unsigned, so this doesn't change any calling code.

## ⚠️ Correction — this does not explain #1011

I originally opened this PR believing it fixed the macOS-arm64 divergence in #1011, based on `--tcp-sequence`'s adjustment value using this function and observing a constant per-packet offset. That was wrong. Direct diagnostic instrumentation run on the actual affected arm64 hardware showed:

- `tcp_sequence_adjust` computes to the **identical** value (`2129613237`) on both macOS arm64 and macOS Intel for seed 42 — this function's output was never actually diverging for this input, before or after this fix.
- The real bug is in the rewrite *application* path, not seed derivation — see #1013 for the actual root cause (a wrong end-of-buffer pointer computed from a stack address instead of the packet buffer, in both `rewrite_sequence.c` and `portmap.c`).

Leaving this PR open on its own merits (it's a real, legitimate UB fix, independently useful) but no longer claiming it resolves #1011.

## Testing

- Confirmed this doesn't change output on macOS Intel (full test suite 69/69, byte-identical to before) — consistent with the UB apparently already resolving to the well-defined-equivalent value on this platform's compiler.
- Confirmed via the arm64 hardware (indirectly, through #1011's diagnosis) that this fix also doesn't change `tcp_sequence_adjust`'s output there — the UB never actually manifested as a different value for this specific seed/iteration count on either platform.

### DIFF
--- a/docs/CHANGELOG
+++ b/docs/CHANGELOG
@@ -7,6 +7,10 @@ Unreleased
       (u_char *)(*ip_hdr)); get_layer4_v4()/v6()'s bounds check against that garbage pointer then
       either always passed (silently correct by luck) or always failed (silently rewriting
       nothing), depending on platform-specific stack layout (#1011)
+    - tcpr_random: fix signed left-shift undefined behavior in this "consistent across all
+      platforms" PRNG - left-shifting a signed int such that a set bit crosses the sign bit is
+      undefined behavior in C, which different compiler backends are free to resolve differently;
+      switched the accumulator to unsigned so the shift is well-defined everywhere
     - tcpliveplay: fix catch_alarm() not calling pcap_breakloop(), which left pcap_dispatch()
       blocked for the full buffer timeout (or indefinitely, on libpcap implementations where the
       timeout only starts once a packet arrives) instead of returning promptly when the alarm
--- a/src/common/utils.c
+++ b/src/common/utils.c
@@ -436,7 +436,18 @@ uint32_t
 tcpr_random(uint32_t *seed)
 {
     unsigned int next = *seed;
-    int result;
+    /*
+     * Must be unsigned: by the third `result <<= 10` below, result can carry
+     * up to ~31 significant bits, close enough to the sign bit that a signed
+     * int here can shift a set bit into (or past) the sign bit - undefined
+     * behavior in C, which different compiler backends are free to resolve
+     * differently. That's exactly what happened: this function produced
+     * different output on macOS arm64 than on macOS Intel/Linux arm64 for
+     * the same seed, despite this function's whole purpose being consistent
+     * output across platforms (see #1011). Unsigned left-shift overflow is
+     * well-defined (mod 2^32 wraparound), which removes the UB entirely.
+     */
+    unsigned int result;
 
     next *= 1103515245;
     next += 12345;
--- a/src/common/utils.c
+++ b/src/common/utils.c
@@ -436,17 +436,6 @@ uint32_t
 tcpr_random(uint32_t *seed)
 {
     unsigned int next = *seed;
-    /*
-     * Must be unsigned: by the third `result <<= 10` below, result can carry
-     * up to ~31 significant bits, close enough to the sign bit that a signed
-     * int here can shift a set bit into (or past) the sign bit - undefined
-     * behavior in C, which different compiler backends are free to resolve
-     * differently. That's exactly what happened: this function produced
-     * different output on macOS arm64 than on macOS Intel/Linux arm64 for
-     * the same seed, despite this function's whole purpose being consistent
-     * output across platforms (see #1011). Unsigned left-shift overflow is
-     * well-defined (mod 2^32 wraparound), which removes the UB entirely.
-     */
     unsigned int result;
 
     next *= 1103515245;


fklassen:
Closing without merging - real UB fix but unrelated to #1011 (confirmed fixed separately by #1013), and not worth carrying as a standalone change right now.

PR #1014, opened 2026-07-17
configure: stop --with-netmap from leaking -I<DIR>/sys into CFLAGS

### LINKED ISSUE #1015: [Bug] --with-netmap build fails with 'redefinition of struct bpf_insn' (--enable-asan on Linux)
## Describe the bug

Configuring with `--with-netmap=DIR` together with `--enable-asan` on Linux fails to build with a `struct bpf_insn` redefinition error.

## Reproduction

```sh
mkdir build && cd build
../autogen.sh
../configure --with-testnic=enp0s5 --with-netmap=/home/fklassen/git/netmap --enable-asan
make
```

## Error

```
Making all in common
  CC       cidr.o
In file included from /usr/include/pcap/pcap.h:133,
                 from /usr/include/pcap.h:43,
                 from ../../src/defines.h:76,
                 from /home/fklassen/git/tcpreplay/src/common/cidr.c:21:
/usr/include/pcap/bpf.h:244:8: error: redefinition of 'struct bpf_insn'
  244 | struct bpf_insn {
      |        ^~~~~~~~
compilation terminated due to -Wfatal-errors.
make[4]: *** [Makefile:481: cidr.o] Error 1
```

Builds fine without `--enable-asan` using the same `--with-netmap`/`--with-testnic` flags.

## Root cause

`configure.ac`'s `have_netmap` detection block permanently appends `-I$NETMAPINCDIR` (the netmap checkout's `sys/` directory) to the global `CFLAGS`, with no restore afterward - unlike the equivalent `CPPFLAGS` addition a few lines later in the same block, which *is* correctly scoped/restored. A netmap checkout's `sys/` tree mirrors a BSD kernel source layout, including its own `sys/net/bpf.h`. With that path permanently in `CFLAGS`, the later `AC_CHECK_HEADERS([net/bpf.h])` feature check (meant to detect a real BSD system header, used for the macOS/BSD BPF injection backend) finds netmap's vendored copy instead - even on Linux, where no such header should exist - and wrongly sets `HAVE_BPF=1`. That activates a `src/defines.h.in` code path meant only for genuine BSD systems (`#include <net/bpf.h>` ahead of the system libpcap headers), and the anti-double-include guard doesn't stop the system libpcap from also defining its own `struct bpf_insn`, producing the conflict.

Full trace and fix: #1014

## System

- OS: Linux (arch not specified by reporter, x86_64 assumed from `enp0s5` interface naming)
- Building against an out-of-tree netmap checkout via `--with-netmap=DIR`
- Fails specifically with `--enable-asan` added to an otherwise-working `--with-netmap`/`--with-testnic` configure line

fklassen:
## Summary

Fixes #1015 — `--with-netmap=DIR --enable-asan` (and, per the root cause, likely any `--with-netmap` build with the right system libpcap) build failure:

```
In file included from /usr/include/pcap/pcap.h:133,
                 from /usr/include/pcap.h:43,
                 from ../../src/defines.h:76,
                 from /home/fklassen/git/tcpreplay/src/common/cidr.c:21:
/usr/include/pcap/bpf.h:244:8: error: redefinition of 'struct bpf_insn'
```

## Root cause

`have_netmap`'s detection block (`configure.ac`) unconditionally appended `-DNETMAP_WITH_LIBS -DND -I$NETMAPINCDIR` to the **global** `CFLAGS` as soon as a netmap checkout was found, with no restore anywhere afterward — unlike the equivalent `CPPFLAGS` addition a few lines later, which *is* correctly scoped (saved to `OLDCPPFLAGS`, restored right after the netmap-version-specific checks that need it). A dead `OLDCFLAGS` assignment sitting right next to the `CFLAGS` mutation suggests a save/restore was intended here too, but never actually implemented — `OLDCFLAGS` was computed but never read again anywhere in the script.

Since `CFLAGS` stayed permanently polluted with the netmap include path for the rest of `configure.ac`, every later `AC_CHECK_HEADERS`/`AC_COMPILE_IFELSE` test (autoconf compiles test programs using `CPPFLAGS` and `CFLAGS` together) could also pick up headers from inside the netmap checkout. A netmap source tree mirrors a BSD kernel source layout under `sys/` — including its own `sys/net/bpf.h` — so the later `checking for net/bpf.h` feature check (meant to detect a genuine BSD system header, used for the macOS/BSD BPF injection backend) found netmap's vendored copy instead and set `have_bpf=yes` even on Linux, where no such header should exist. That wrongly defines `HAVE_BPF`, which activates a `defines.h.in` code path meant only for real BSD systems (`#include <net/bpf.h>`) ahead of the system libpcap headers — and on the reporter's system, the anti-double-include guard (`PCAP_DONT_INCLUDE_PCAP_BPF_H`) didn't stop the system libpcap from also defining its own copy of `struct bpf_insn`, producing the conflict.

## Fix

Don't touch the shared `CFLAGS` during netmap detection. Save the netmap include flags into a dedicated `NETMAP_CFLAGS` variable instead, and apply it to `CFLAGS` exactly once, right before `AC_CONFIG_FILES`/`AC_OUTPUT` — after every other `AC_CHECK_*`/`AC_COMPILE_IFELSE` test in the script has already run against a clean `CFLAGS`. The real build (where `netmap.c` etc. need to find netmap's headers) still gets the same `CFLAGS` addition, just deferred to the point where it can no longer shadow unrelated system headers during configure's own detection.

## Verification

**This dev environment has no netmap checkout**, so I cannot reproduce or fully verify the `--with-netmap` path directly — flagging clearly. What I could verify:

- Confirmed the non-netmap path (the common case) is completely unaffected: `NETMAP_CFLAGS` stays empty, the new `if` block is a no-op, full rebuild and local test suite (69/69) unchanged.
- Regenerated `configure` via `autogen.sh` cleanly — valid autoconf/m4 syntax, only the usual pre-existing `AC_HEADER_STDC` obsolete warning (unrelated).
- Traced the exact mechanism precisely against the reporter's error trace and their `--with-netmap`/`--enable-asan` configure command — the line numbers, header chain, and `struct bpf_insn` collision all line up exactly.

Given I can't test the netmap path myself, would very much appreciate a run of `../configure --with-testnic=enp0s5 --with-netmap=<your netmap checkout> --enable-asan && make` on the reporter's original setup before merge.

## Test plan

- [x] `autogen.sh` regenerates `configure` cleanly
- [x] Non-netmap configure + full build + local test suite: 69/69, no regression
- [ ] Confirm on the original Linux + netmap + `--enable-asan` setup that the build now succeeds

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1344,7 +1344,14 @@ for testdir in $NETMAP_SEARCH_DIRS; do
         dnl redefinition conflict against the system libpcap headers at build
         dnl time. Save it in NETMAP_CFLAGS and apply it to CFLAGS just once,
         dnl right before AC_OUTPUT, after all other detection has finished.
-        NETMAP_CFLAGS="-DNETMAP_WITH_LIBS -DND -I$NETMAPINCDIR"
+        dnl
+        dnl Deliberately NOT passing -DND: netmap_user.h defines its own
+        dnl ND(...) debug-print macro guarded by #ifndef ND. A bare -DND (no
+        dnl value) makes the compiler define ND as the literal 1, so that
+        dnl guard sees ND as already defined and skips its own function-like
+        dnl macro - every ND(...) call site then expands to 1(...), "call
+        dnl the integer 1 as a function", a hard compile error (#1015).
+        NETMAP_CFLAGS="-DNETMAP_WITH_LIBS -I$NETMAPINCDIR"
 
         AC_SUBST(NETMAPINC)
         AC_SUBST(NETMAPUSERINC)
@@ -1369,7 +1376,8 @@ dnl ###########################################################
 
 if test "x$NETMAPINCDIR" != "x"; then
     OLDCPPFLAGS="$CPPFLAGS"
-    CPPFLAGS="$CPPFLAGS -DNETMAP_WITH_LIBS -DND -I$NETMAPINCDIR"
+    dnl No -DND here either - see the NETMAP_CFLAGS comment above (#1015).
+    CPPFLAGS="$CPPFLAGS -DNETMAP_WITH_LIBS -I$NETMAPINCDIR"
 
     have_nm_open=no
     have_nm_nr_reg_mask=no
--- a/configure.ac
+++ b/configure.ac
@@ -1334,8 +1334,17 @@ for testdir in $NETMAP_SEARCH_DIRS; do
 
     if test "$have_netmap" != no ; then
         NETMAPFLAGS="-DHAVE_NETMAP"
-        OLDCFLAGS="$CFLAGS -I$NETMAPINCDIR"
-        CFLAGS="$CFLAGS -DNETMAP_WITH_LIBS -DND -I$NETMAPINCDIR"
+        dnl Don't add -I$NETMAPINCDIR to CFLAGS here: CFLAGS is shared by every
+        dnl AC_CHECK_HEADERS/AC_COMPILE_IFELSE test for the rest of this script,
+        dnl and a netmap checkout's sys/ tree mirrors a BSD kernel source layout
+        dnl (including its own sys/net/bpf.h). Adding it now makes later,
+        dnl unrelated checks - e.g. the net/bpf.h check below - find netmap's
+        dnl vendored BSD-compat header instead of a real system one, wrongly
+        dnl setting have_bpf=yes even on Linux and causing a struct bpf_insn
+        dnl redefinition conflict against the system libpcap headers at build
+        dnl time. Save it in NETMAP_CFLAGS and apply it to CFLAGS just once,
+        dnl right before AC_OUTPUT, after all other detection has finished.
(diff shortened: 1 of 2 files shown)


fklassen:
Found and fixed a second, related bug while waiting on real-hardware verification of the CFLAGS leak fix above: `--with-netmap=DIR` also passed a bare `-DND` in both `NETMAP_CFLAGS` and `CPPFLAGS`. netmap's own `netmap_user.h` defines a function-like `ND(...)` debug macro guarded by `#ifndef ND` — a valueless `-DND` predefines `ND` as `1`, so every real `ND(...)` call site in netmap's headers expands to `1(...)`, i.e. "called object is not a function or function pointer".

Root-caused from the build error pasted against `../configure --with-testnic=enp0s5 --with-netmap=/home/fklassen/git/netmap`. Fix: drop `-DND` from both spots, same as the CFLAGS-leak fix above.

Same disclosure as before: I have no netmap checkout or Linux box, so I've only verified the non-netmap path still builds clean and `sudo make test` is 69/69 with no regressions. The actual `--with-netmap` + `--enable-asan` combination still needs your hardware to confirm.

PR #1061, opened 2026-07-22
fix(txring): release resources on close

### LINKED ISSUE #1059: tcpreplay: TX_RING resources (thread, mmap, 2 allocations) leaked on every run under ASan
Reported via ASan (\`--enable-asan\`) while running \`sudo make test\`:

```
==153032==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 in malloc
    #1 in our_safe_malloc ../../../src/common/utils.c:42
    #2 in txring_init ../../../src/common/txring.c:200
    #3 in sendpacket_open_pf ../../../src/common/sendpacket.c:1332
    #4 in sendpacket_open ../../../src/common/sendpacket.c:760
    #5 in tcpreplay_post_args ../../src/tcpreplay_api.c:424
    #6 in main ../../src/tcpreplay.c:71

SUMMARY: AddressSanitizer: 64 byte(s) leaked in 2 allocation(s).
```

**Root cause**: `txring_init()` (`src/common/txring.c`) allocates a `txring_t` (`txp`) and its `tpacket_req` (`txp->treq`), `mmap()`s the TX ring buffer, and spawns a poll thread (`txring_send()`) that loops `while (!shutdown_flag)`. There is no `txring_close()` anywhere in the codebase — `sendpacket_close()`'s `SP_TYPE_TX_RING` case just does `close(sp->handle.fd)`. Neither allocation is ever freed, the mmap is never unmapped, and the poll thread is never signaled to stop or joined — it runs until the process exits. The `shutdown_flag` global that the thread loop checks is set to 0 at startup and never written anywhere else in the tree, so it's dead code as it stands today.

This means every `tcpreplay`/`tcpreplay-edit` invocation that opens a PF_PACKET/TX_RING interface (the default injection method on Linux) leaks these resources at exit, and — more importantly for `libtcpreplay` users — leaks a live background thread + mmap for every interface opened within a single long-running process, not just at final process exit.

Fix in progress: give `txring_t` a per-instance `shutdown_flag` (the existing global would incorrectly stop *all* open TX rings if any one of them is closed, e.g. in dual-NIC replay), add `txring_close()` (set the flag, `pthread_join()`, `munmap()`, free both allocations), and call it from `sendpacket_close()`'s `SP_TYPE_TX_RING` case before closing the fd. Also fixes the same leak on `txring_init()`'s existing error-return paths (setsockopt/mmap failure), which leaked `txp`/`txp->treq` even before this fix.

Solaris-star:
## Summary

`txring_init()` allocates a TX_RING context, request structure, mmap, and sender thread, but the existing close path only closes the socket. This leaks resources and leaves the sender thread running for library users that open and close interfaces repeatedly.

## Fix

- Add per-instance TX_RING shutdown state instead of the shared global flag.
- Pass the ring context to the sender thread so multiple TX_RING instances can be closed independently.
- Add `txring_close()` to stop/join the thread, unmap the ring, disable the socket ring, and free allocations.
- Call `txring_close()` from `sendpacket_close()` before closing the socket.
- Clean up allocated state and enabled socket rings on initialization failures.
- Destroy pthread attributes after thread creation attempts.

Fixes #1059

## Verification

- [x] `git diff --check`
- [x] Full tracked-file sensitive-data scan
- [ ] Linux autotools/CI build (the local host is macOS; the project CI is Linux-only)

### DIFF
--- a/src/common/sendpacket.c
+++ b/src/common/sendpacket.c
@@ -767,8 +767,14 @@ sendpacket_close(sendpacket_t *sp)
         break;
 
     case SP_TYPE_PF_PACKET:
-    case SP_TYPE_TX_RING:
 #ifdef HAVE_PF_PACKET
+        close(sp->handle.fd);
+#endif
+        break;
+
+    case SP_TYPE_TX_RING:
+#ifdef HAVE_TX_RING
+        txring_close(sp->tx_ring);
         close(sp->handle.fd);
 #endif
         break;
--- a/src/common/txring.c
+++ b/src/common/txring.c
@@ -44,7 +44,6 @@
 #include <sys/mman.h>
 #include <unistd.h>
 
-volatile int shutdown_flag = 0;
 int tdata_offset = TPACKET_HDRLEN - sizeof(struct sockaddr_ll);
 
 /**
@@ -54,12 +53,12 @@ void *
 txring_send(void *arg)
 {
     int ec_send;
-    static int total = 0;
-    int fd_socket = (int)arg;
+    int total = 0;
+    txring_t *txp = arg;
 
     do {
         /* send all buffers with TP_STATUS_SEND_REQUEST */
-        ec_send = sendto(fd_socket, NULL, 0, MSG_DONTWAIT, (struct sockaddr *)NULL, sizeof(struct sockaddr_ll));
+        ec_send = sendto(txp->fd, NULL, 0, MSG_DONTWAIT, (struct sockaddr *)NULL, sizeof(struct sockaddr_ll));
 
         if (ec_send > 0) {
             total += ec_send;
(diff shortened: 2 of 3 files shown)


fklassen:
Thanks for this — you independently found and fixed the same leak reported in #1059, which landed via #1060 shortly before this PR's base got retargeted, hence the conflicts.

Rather than have you rebase over already-merged, overlapping code, I picked out the two things your version does better than what's already merged — the `pthread_attr_destroy()` you added (a real leak neither the original code nor #1060 caught) and explicitly disabling `PACKET_TX_RING` via `setsockopt` before unmapping — and shipped them as a credited follow-up: #1063.

Closing this one as superseded, but appreciated the independent find and the cleanup ideas.

PR #1066, opened 2026-07-23
fragroute: fix off-by-one in tcp_chaff and unchecked packet allocations

fklassen:
Follow-up to the fragroute security work on `4.6.0-beta2`. Both issues here were found while auditing the fix for GHSA-m655-53p4-6qm8, and neither is covered by any of the three advisories.

## 1. Off-by-one heap overflow in `tcp_chaff` (security)

`tcp_chaff_apply()` had the identical defect that GHSA-m655-53p4-6qm8 reported in `ip_chaff`:

```c
rand_strset(data->rnd, new->pkt_tcp_data, new->pkt_end - new->pkt_tcp_data + 1);
```

The writable span is `[pkt_tcp_data, pkt_end)`, so the `+ 1` writes one byte past the end of the allocation `pkt_dup()` just returned. Any `tcp_chaff` directive against an ordinary TCP packet triggers it — no crafted packet needed, same trust boundary as the reported bugs (an attacker-influenced `--fragroute=` rules file).

Confirmed under AddressSanitizer, with a signature matching the `ip_chaff` one exactly:

```
ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 1
    #0 in rand_strset      src/fragroute/randutil.c:58
    #1 in tcp_chaff_apply  src/fragroute/mod_tcp_chaff.c:107
0x50800000007f is located 0 bytes after 95-byte region
    #1 in pkt_dup          src/fragroute/pkt.c:72
```

## 2. Unchecked `pkt_new()` / `pkt_dup()` (robustness)

Six call sites dereferenced the returned packet immediately, so an allocation failure was a NULL dereference rather than an aborted rule: two each in `ip_frag_apply_ipv4()` and `ip_frag_apply_ipv6()`, one plus its dup in `tcp_seg_apply()`, and the one in `tcp_chaff_apply()`. `mod_dup.c` already handled this correctly and is unchanged.

## Verification

ASan build of this branch:

- `tcp_chaff cksum` and `tcp_chaff null` — clean (both reproduced the overflow before the fix)
- The five inputs fixed in the previous commit — still clean
- Ten valid rule sets (`ip_frag`/`tcp_seg` in both overlap modes, `tcp_chaff`, `ip_chaff`, `dup`, `order`, `ip_ttl`) — still produce output, no sanitizer reports

## Notes

- No advisory exists for the `tcp_chaff` overflow. Given it's the same class, severity and trust boundary as GHSA-m655-53p4-6qm8, it probably warrants one, credited as found during remediation rather than to the original reporter.
- `test/Makefile.am` still has no fragroute targets, so none of this is covered by `make test`. Test cases are still outstanding.
- Not addressed here: `tmp_buf[pkt->pkt_buf_size]` in `mod_ip_frag.c` (x2) and `mod_tcp_seg.c` is a stack VLA sized from packet-derived data. Worth bounding, but it's a more invasive change than this PR's scope.

https://claude.ai/code/session_01W2Rdq79uuiVVq9G9GyuvDw


fklassen:
Added fragroute test cases (`4b628de5`) — the gap noted in the original description.

Three targets in the `tcprewrite` group, gated on `COMPILE_FRAGROUTE`:

| target | covers |
| --- | --- |
| `fragroute_valid` | 12 valid rule sets across `ip_frag`, `tcp_seg`, `ip_chaff`, `tcp_chaff`, `dup`, `order`, `ip_ttl` |
| `fragroute_negsize` | negative/zero sizes refused at parse time, without crashing |
| `fragroute_badrules` | empty, comment-only, and 300-directive rules files |

There is no golden file to diff against — fragroute embeds random IP IDs and random chaff bytes, so output is not reproducible run to run. The targets check exit status instead, distinguishing a clean rejection (tcprewrite exits 255 on a fatal error) from a crash (128..254 means a signal).

**One limitation worth being explicit about.** I verified these against the pre-fix tree. `fragroute_negsize` and `fragroute_badrules` both fail there with SIGSEGV and pass on this branch. `fragroute_valid` **passes against the buggy code** — a one-byte heap overflow doesn't crash an unsanitized binary, so it cannot catch the `ip_chaff`/`tcp_chaff` off-by-ones in an ordinary build. It needs `--enable-asan` to catch those. That's documented in a comment above the target.

If you want those regressions caught automatically, CI would need to run `make test` against an ASan build; right now nothing does.

`test.pcap` was unusable here — fragroute treats the first non-IP packet as fatal and 109 of its 259 packets are non-IP. `test_fragroute.pcap` is a 20-packet IP-only subset (5 KB), 15 of them TCP with ACK and payload so `tcp_seg`/`tcp_chaff` have something to act on.

